PR #67, opened 2022-01-05
Basic connection

Choraden:


mmatczuk:
@Choraden please port your work to https://github.com/mmatczuk/scylla-go-driver/pull/88.
There is an integration test added it needs to be extended and connection needs to be extended too.

Choraden:
I don't quite understand which parts of my work you want to be ported. 

>There is an integration test added it needs to be extended

Do you want me to extend integration test to send simple query?

What's the form? Should I make a PR to #88?

mmatczuk:
I want to port the query test. I needs query request support.
You can send PR against #88, I hope we can merge it soonish so the you can send to main.

mmatczuk:
Let's close that.

PR #88, opened 2022-01-14
Connection infra

mmatczuk:
This is WIP, it adds connection foundation for connection handling


frame/header.go
```diff
@@ -1,37 +1,34 @@
 package frame
 
-import (
-	"fmt"
-)
+// StreamID is a type alias for Short.
+type StreamID = Short
```
Choraden, line 4: All the types are currently declared in types.go, I would consider moving it there.

transport/conn.go
```diff
@@ -0,0 +1,222 @@
...
+	h.WriteTo(&c.buf)
+	r.WriteTo(&c.buf)
+	if err := c.buf.Error(); err != nil {
+		return fmt.Errorf("serialize: %w", err)
+	}
+
+	// Update length in header
+	b := c.buf.Bytes()
+	l := uint32(len(b) - frame.HeaderSize)
+	binary.BigEndian.PutUint32(b[5:9], l)
```
Choraden, line 72: How about Buffer method to update header length?

transport/conn.go
```diff
@@ -0,0 +1,222 @@
...
+
+func WrapConn(conn net.Conn, s StreamIDAllocator) *Conn {
+	c := &Conn{
+		conn: conn,
+		w: connWriter{
+			conn:      conn,
+			requestCh: make(chan request, 1024),
+		},
+		r: connReader{
+			conn: bufio.NewReaderSize(conn, 8192),
```
Neofine, line 186: As above.

transport/conn.go
mmatczuk: Buffer operations are twofold, read new bytes from a stream and write (add) bytes to slice.
This is different and requires some additional assumptions.
This could be moved to a separate function in frame pkg if needed - then it can be unit tested.

transport/conn.go
```diff
@@ -0,0 +1,222 @@
...
+
+	stream StreamIDAllocator
+}
+
+func WrapConn(conn net.Conn, s StreamIDAllocator) *Conn {
+	c := &Conn{
+		conn: conn,
+		w: connWriter{
+			conn:      conn,
+			requestCh: make(chan request, 1024),
```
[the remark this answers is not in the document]
mmatczuk: 👍

transport/conn.go
```diff
@@ -0,0 +1,229 @@
...
+	// Read header
+	if _, err := io.CopyN(c.bufw, c.conn, frame.HeaderSize); err != nil {
+		r.Err = fmt.Errorf("read header: %w", err)
+		return r
+	}
+	r.Header = frame.ParseHeader(&c.buf)
+	if err := c.buf.Error(); err != nil {
+		r.Err = fmt.Errorf("parse header: %w", err)
+		return r
+	}
```
Kulezi, line 141: If we adopt changes from #103 these parts should be able to resolve panics

transport/conn.go
```diff
@@ -0,0 +1,229 @@
...
+	// Read body
+	if _, err := io.CopyN(c.bufw, c.conn, int64(r.Header.Length)); err != nil {
+		r.Err = fmt.Errorf("read body: %w", err)
+		return r
+	}
+	r.Response = c.parse(r.Header.OpCode)
+	if err := c.buf.Error(); err != nil {
+		r.Err = fmt.Errorf("parse body: %w", err)
+		return r
+	}
```
Kulezi, line 152: same as above

transport/conn.go
```diff
@@ -0,0 +1,229 @@
...
+	h := frame.Header{
+		Version:  frame.CQLv4,
+		StreamID: r.StreamID,
+		OpCode:   r.OpCode(),
+	}
+	h.WriteTo(&c.buf)
+	r.WriteTo(&c.buf)
+	if err := c.buf.Error(); err != nil {
+		return fmt.Errorf("serialize: %w", err)
+	}
```
Kulezi, line 75: This should always succeed if we adopt changes in #103

Michal-Leszczynski (approved):

Issue #70, opened 2022-01-05
frame: use var for frame.Buffer initialization


mmatczuk:
Replace `b := frame.Buffer{}` with var equivalent.

Issue #56, opened 2021-12-16
move ErrorCode constants to top level pkg


mmatczuk:
Move the errors to errors.go in the repo root.
Framer should not try to validate the error code.
The error codes are for driver users to get to know what the error means.

It would be best to have proper errors that can print themselves with descriptions but that we can do later.

Issue #50, opened 2021-12-16
frame: ReadValue use of magic constants


mmatczuk:
```
// Length equal to -1 represents null.
// Length equal to -2 represents not set.
func (b *Buffer) ReadValue() Value {
	if n := b.ReadInt(); n < -2 {
		b.RecordError(fmt.Errorf("invalid value length"))
	} else if n > 0 {
		return Value{N: n, Bytes: b.Read(int(n))}
	} else {
		return Value{N: n}
	}
	return Value{}
}
```

Get rid of the comment and add special Values, ex

```
var NotSetValue = Value{N: -2}
```

Issue #47, opened 2021-12-16
frame: type demarcations use Byte not byte


mmatczuk:
Example 

```
type BatchTypeFlag = byte
```

PR #91, opened 2022-01-14
frame: bugs and little missed out things fixed

Neofine:
- Changed to var for frame.Buffer initialization in skipped places (#70)
- Moved error codes to top level package, out of frame (#56)
- ReadValue no longer uses magic constants (#50)
- Changed type demarcations from byte to Byte (#47)

### DIFF
--- a/frame/buffer.go
+++ b/frame/buffer.go
@@ -3,11 +3,13 @@ package frame
 import (
 	"bytes"
 	"fmt"
+	"log"
 )
+
+var Debug = true
 
 type Buffer struct {
 	buf bytes.Buffer
-	err error
 }
 
 func (b *Buffer) Bytes() []byte {
@@ -18,39 +20,15 @@ func (b *Buffer) Reset() {
 	b.buf.Reset()
 }
 
-func (b *Buffer) Error() error {
-	return b.err
-}
-
-func (b *Buffer) recordError(err error) {
-	if b.err != nil {
-		return
-	}
-
-	b.err = err
-}
-
 func (b *Buffer) Write(v Bytes) {
-	if b.err != nil {
-		return
-	}
-
 	_, _ = b.buf.Write(v)
 }
 
 func (b *Buffer) WriteByte(v Byte) {
-	if b.err != nil {
-		return
-	}
-
 	_ = b.buf.WriteByte(v)
 }
 
 func (b *Buffer) WriteShort(v Short) {
-	if b.err != nil {
-		return
-	}
-
 	_, _ = b.buf.Write([]byte{
 		byte(v >> 8),
 		byte(v),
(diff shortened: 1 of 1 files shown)


Neofine:
@mmatczuk I have closed previous PR fixing those things (https://github.com/mmatczuk/scylla-go-driver/pull/76) because of overwhelm of conflicts inflicted after some actions on main branch. This is the new one.

errors/errors.go
```diff
@@ -0,0 +1,80 @@
+package errors
```
mmatczuk, line 1: Errors is not a good pkg name because it is often use, for instance there is very popular pkg/errors.
So you would endup having conflicts and there is not contextual information
Let's put that to the repo root.

mmatczuk (changes requested):
Please try to fix different issues in different commits, otherwise it's very confusing in review.

If issue states put sth to root and you want to suggest a better approach let's discuss that first.

The commit message must be in appropriate form to automatically close issues, see the [spec](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue), basically add `Fixes #XX` as last line of a commit. You can put many such lines.

errors/errors.go
Neofine: > If issue states put sth to root and you want to suggest a better approach let's discuss that first.

My apologies, forgot to mention this, the linter was quite upset when I put errors.go in root directory
because I would need to import 'scylla-go-driver' and it doesn't like dashes in import, we couldn't
find fix for that so that was my temporary band-aid solution to this.

Neofine:
> Please try to fix different issues in different commits, otherwise it's very confusing in review.

@mmatczuk Should I make new branch and fix my mistake? I don't really know how severe my mistake is.

mmatczuk:
- Do not do that in future.
- Fix commit msg.
- Fix code.

frame/response/error.go
```diff
@@ -1,13 +1,14 @@
...
 import (
+	scyllagodriver "scylla-go-driver"
 	"scylla-go-driver/frame"
 )
 
 // Error response message type used in non specified errors which don't have a body.
 // Error spec: https://github.com/apache/cassandra/blob/trunk/doc/native_protocol_v4.spec#L1046
 type Error struct {
-	Code    frame.ErrorCode
+	Code    scyllagodriver.ErrorCode
```
mmatczuk, line 11: Let's use frame.Int to avoid the dependency.

frame/response/error_test.go
```diff
@@ -1,6 +1,7 @@
 package response
 
 import (
+	scyllagodriver "scylla-go-driver"
```
mmatczuk, line 4: Maybe just scylla

mmatczuk:
@Neofine some minor items.

mmatczuk:
You cannot import scylla pkg from frame because it will create a dependency loop.
I commented on that `Let's use frame.Int to avoid the dependency.`.

Neofine:
> You cannot import scylla pkg from frame because it will create a dependency loop.

Apologies, I thought you meant only error.go file, I will fix that right away!

PR #94, opened 2022-01-15
frame: buffer read, improved error handling and made method private

### LINKED ISSUE #82: frame: buffer do not export Read function
Clients should only use the ReadSth functions.
There are no usages except for buffer itself.

### LINKED ISSUE #83: frame: buffer Read improve error handling
We have 

```
	l, err := b.buf.Read(p)
	if l != n || err != nil {
		b.recordError(fmt.Errorf("invalid Buffer Read"))
	}
```

We want 
- Invalid len to be reported separately 
- Original error to be recorded

Choraden:
Fixes #82
Fixes #83

### DIFF
--- a/frame/buffer.go
+++ b/frame/buffer.go
@@ -332,31 +332,36 @@ func (b *Buffer) WriteQueryOptions(q QueryOptions) { // nolint:gocritic
 	}
 }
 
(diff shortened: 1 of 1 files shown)


Choraden:
Buffer does not use readRawBytes method, but I left it in case we would like to use it in the future.

Issue #71, opened 2022-01-10
frame: optimize huge params 


Choraden:
Consider passing huge params by pointer.

mmatczuk:
This it resolved.

PR #95, opened 2022-01-15
Pointers issue

Neofine:


frame/request/batch.go
```diff
@@ -51,7 +51,7 @@ type BatchQuery struct {
 	Values   []frame.Value
 }
 
-func (q BatchQuery) WriteTo(b *frame.Buffer, name bool) { // nolint:gocritic
+func (q *BatchQuery) WriteTo(b *frame.Buffer, name bool) { // nolint:gocritic
```
Choraden, line 54: Passing by pointer should resolve gocritic issue (#71) as well, so nolint comments can be deleted.

Neofine: Great! Will change it right away, thx!

Choraden: Actually there is one more gocritic nolint in buffer.go so don't link this issue until this is fixed.

Neofine:
I am closing this PR as some refactoring was done.

Issue #98, opened 2022-01-18
frame: error reporting considered futile


mmatczuk:
After a closer examination I think that the buffer error reporting is useless.

Write path - it will never happen that we send invalid data.
```
WriteConsistency
		b.recordError(fmt.Errorf("invalid consistency: %v", v))
WriteEventTypes
			b.recordError(fmt.Errorf("invalid EventType %s", k))
WriteInet
		b.recordError(fmt.Errorf("invalid IP length"))
WriteOpCode
		b.recordError(fmt.Errorf("invalid operation code: %v", v))
WriteStartupOptions
			b.recordError(fmt.Errorf("invalid mandatory Startup option %s: %s", k, s))
			b.recordError(fmt.Errorf("invalid Startup option %s: %s", k, s))
		b.recordError(fmt.Errorf("invalid Startup option"))
WriteValue
		b.recordError(fmt.Errorf("invalid value"))
```

To save this work for future I suggest adding `frame.Debug` variable.
It the flag is on we check the condition, if condition is met we log it.

Read path - premature handling of unsupported options.

```
ReadConsistency
		b.recordError(fmt.Errorf("invalid consistency: %v", v))
ReadErrorCode
		b.recordError(fmt.Errorf("invalid error code: %d", v))
ReadInet
	b.recordError(fmt.Errorf("invalid ip length"))
ReadOpCode
		b.recordError(fmt.Errorf("invalid operation code: %v", o))
ReadOption
			b.recordError(fmt.Errorf("invalid Option ID: %d", id))
ReadSchemaChangeTarget
		b.recordError(fmt.Errorf("invalid SchemaChangeTarget: %s", v))
ReadSchemaChangeType
		b.recordError(fmt.Errorf("invalid SchemaChangeType: %s", t))
ReadStatusChangeType
		b.recordError(fmt.Errorf("invalid StatusChangeType: %s", t))
ReadTopologyChangeType
		b.recordError(fmt.Errorf("invalid TopologyChangeType: %s", t))
ReadValue
		b.recordError(fmt.Errorf("invalid value length"))
ReadWriteType
		b.recordError(fmt.Errorf("invalid write type: %s", w))
```

If server sent OpCode we do not support the place to handle that is in the main parsing function.
The fact that constant exists does not guarantee that it is supported anyways.
Checking for those errors actually is mixed up with i/o errors wich is unfortunate.
I suggest we do the same as with Write path - use `frame.Debug` variable.

[...]

PR #100, opened 2022-01-18
Frame: buffer, optimized UUID reading.

### LINKED ISSUE #80: frame: optimize reading UUID
Example

```
	var x [16]byte
	var v bytes.Buffer
	v.Read(x[:])
```

Michal-Leszczynski:
- optimized UUID reading
- added simple test for UUID
Fixes #80


frame/buffer.go
```diff
@@ -427,12 +427,11 @@ func (b *Buffer) ReadOpCode() OpCode {
...
 	if b.err != nil {
-		return UUID{0}
+		return UUID{}
 	}
 
-	s := b.read(16)
-	u := UUID{}
-	copy(u[:], s)
+	var u UUID
+	_, b.err = b.buf.Read(u[:])
```
mmatczuk, line 434: `b.err` shall be set with recordError and wrapped with some context information.

This may change with #98 but for now this is the way to do that.

mmatczuk:
Blocker by #98

mmatczuk (changes requested):
.

frame/buffer.go
Michal-Leszczynski: I changed it to error handling from #98

mmatczuk (approved):
👍

PR #103, opened 2022-01-19
frame: buffer - debug mode dependent error reporting

### LINKED ISSUE #98: frame: error reporting considered futile

Kulezi:
- Introduced frame.Debug flag, setting it to false disables error logging.
- Reading/Writing things such as opcodes that are not expected by the driver
  now produces logs of such events only if frame.Debug flag is set.
- Not being able to read enough data from buffer now forces a panic().

Fixes #98


mmatczuk:
Can we have 

```
if Debug {
   other ifs
}
```

in all places? It makes it easier to read.

mmatczuk:
Leave the header, the header validation is removed in #88

mmatczuk:
Nice job.

Kulezi:
> Leave the header, the header validation is removed in #88

unstaged, it won't pass the check due to obvious reasons of not compiling



> Can we have
> 
> ```
> if Debug {
>    other ifs
> }
> ```
> 
> in all places? It makes it easier to read.

Now it's like this in the whole file.

mmatczuk:
Please fix the linter.

PR #110, opened 2022-01-22
Transport: connection, query capabilities

Choraden:


frame/request/query.go
```diff
@@ -16,6 +16,7 @@ type Query struct {
 func (q *Query) WriteTo(b *frame.Buffer) {
```
mmatczuk, line 16: Cheers for clarifications in the commit message! 
Can you please add one line separator between header and body.
The reason is that Goland is stupid 

image

transport/conn.go
```diff
@@ -158,26 +158,52 @@ func (c *connReader) parse(op frame.OpCode) frame.Response {
...
+		return ParseSetKeyspaceResult(&c.buf)
+	case PreparedKind:
+		return ParsePreparedResult(&c.buf)
+	case SchemaChangeKind:
+		return ParseSchemaChangeResult(&c.buf)
+	default:
+		return nil
+	}
+}
+
```
mmatczuk, line 186: Let's move that functionality to result.go

transport/conn.go
```diff
@@ -158,26 +158,52 @@ func (c *connReader) parse(op frame.OpCode) frame.Response {
...
 type Conn struct {
 	conn net.Conn
 	w    connWriter
 	r    connReader
 
+	cfg ConnConfig
```
mmatczuk, line 192: To me this should be the top most element of this struct.

transport/conn.go
```diff
@@ -188,12 +214,18 @@ func WrapConn(conn net.Conn, s StreamIDAllocator) *Conn {
...
 		},
+		cfg:    cfg,
 		stream: s,
 	}
 	go c.w.loop()
 	go c.r.loop()
 
-	return c
+	if err := c.initializeConnection(); err != nil {
+		// TODO close connection
```
mmatczuk, line 224: Go for it. This should be documented and Close function shall be added.

transport/conn.go
```diff
@@ -202,6 +234,61 @@ func (c *Conn) Startup(options frame.StartupOptions) (frame.Response, error) {
...
+
+	res, err := c.sendRequest(req, s.Compression, s.Tracing)
+	if err != nil {
+		return QueryResult{}, err
+	}
+
+	return intoQueryResult(res)
+}
+
+func intoQueryResult(res frame.Response) (QueryResult, error) {
```
mmatczuk, line 275: This should be called makeQueryResult.
I'd suggest to create a query.go and put all query related structs and helper functions / constructors there.

transport/conn.go
```diff
@@ -202,6 +234,61 @@ func (c *Conn) Startup(options frame.StartupOptions) (frame.Response, error) {
...
+	req := &Query{
+		Query:       s.Content,
+		Consistency: s.Consistency,
+		Options: frame.QueryOptions{
+			Values:            s.Values,
+			SerialConsistency: s.SerialConsistency,
+			PagingState:       pagingState,
+			PageSize:          s.PageSize,
+		},
+	}
```
mmatczuk, line 265: We would benefit from having `newQueryFromStatement` function.

transport/conn.go
```diff
@@ -202,6 +234,61 @@ func (c *Conn) Startup(options frame.StartupOptions) (frame.Response, error) {
...
 
+type QueryResult struct {
+	Rows        []frame.Row
+	Warnings    []string
+	TracingID   frame.UUID
+	PagingState frame.Bytes
+	ColSpec     []frame.ColumnSpec
+}
+
+type statement struct {
```
mmatczuk, line 245: This must be exported if Query is exported.

transport/conn.go
```diff
@@ -158,26 +158,52 @@ func (c *connReader) parse(op frame.OpCode) frame.Response {
...
 	w    connWriter
 	r    connReader
 
+	cfg ConnConfig
+
 	stream StreamIDAllocator
 }
 
+type ConnConfig struct {
+	opts frame.StartupOptions
```
mmatczuk, line 198: What's that?

mmatczuk: We need proper creation of StartupOptions.
This is beyond the scope, add fixed value where needed.

transport/conn_integration_test.go
```diff
@@ -7,21 +7,116 @@ import (
...
+	if _, err := conn.Query(insert2, nil); err != nil {
+		return err
+	}
 
-	opts := frame.StartupOptions{
-		"CQL_VERSION": "3.0.0",
+	return nil
+}
+
+func TestMassiveQueryOneConnection(t *testing.T) {
```
mmatczuk, line 71: I'd suggest TestConnMassiveQueryIntegration

transport/conn_integration_test.go
```diff
@@ -7,21 +7,116 @@ import (
...
+	_, err = WrapConn(nc,
+		TestStreamIDAllocator{},
+		ConnConfig{opts: frame.StartupOptions{"CQL_VERSION": "3.0.0"}},
+	)
+	if err != nil {
+		t.Fatal(err)
+	}
+}
+
+func insertSampleData(conn *Conn) error {
```
mmatczuk, line 28: This can be moved inside the test

mmatczuk (changes requested):
See comments inline.

transport/conn.go
Choraden: Startup options are neccessary to init connection. I thought if it would be better to pass opts as a parameter to init() and in consequence to WrapConn(...), but finally decided to place it in config, because many connections will probably have the same configs. 
ConnConfig will be filled with other variables, but for now I needed only options.

transport/conn.go
Choraden: But do we still want to initialize connection in WrapConn?

Choraden:
TestConnMassiveQueryIntegration needs valid streamId allocator to work properly, since TestStreamIDAllocator makes collisions in streamIDs and blocks driver.

mmatczuk:
Let's merge https://github.com/mmatczuk/scylla-go-driver/pull/111 first.

mmatczuk:
Please adjust to new streamID code.

Choraden:
```TestConnMassiveQueryIntegration``` fails for ``` n > 10000```.
It's the scylla responding with Error.  
Report says:
``` conn_integration_test.go:102: make query result: Operation timed out for mykeyspace.users - received only 0 responses from 1 CL=ONE.```

Maybe it's a matter of my laptop or maybe the driver is too fast :)

PR #111, opened 2022-01-23
transport: add StreamIDAllocator

### LINKED ISSUE #108: Add StreamIDAllocator impl
StreamIDAllocator is specified as

```
type StreamIDAllocator interface {
	Alloc() (frame.StreamID, error)
	Free(id frame.StreamID)
}
```

Gocql has an [implementation](https://github.com/gocql/gocql/blob/master/internal/streams/streams.go) based on buckets with atomics.
Rust uses simpler approach (also with buckets) but under a lock.

I'd try:
- Move getting the ID to `func (c *connReader) setHandler(streamID frame.StreamID, h responseHandler)` it would generate the ID under a lock but we need to take the lock anyways and the scope is only one connection.
- Then we can have simpler implementation like in our Rust driver

Kulezi:
- moved getting streamID from sendRequest() to setHandler(), as a result
    setHandler() changed its return value to (frame.StreamId, err)
- moved stream from Conn to connReader so it's visible in sendRequest()
- added DefaultStreamIDAllocator ported directly from the rust driver
- removed TestStreamIDAllocator, replaced it with the new one in tests.

Fixes #108

### DIFF
--- a/transport/conn.go
+++ b/transport/conn.go
@@ -194,7 +194,6 @@ func WrapConn(conn net.Conn) *Conn {
 		r: connReader{
 			conn: bufio.NewReaderSize(conn, ioBufferSize),
 			h:    make(map[frame.StreamID]responseHandler),
-			s:    streamIDAllocator{},
 		},
 	}
 	go c.w.loop()
@@ -214,7 +213,7 @@ func (c *Conn) sendRequest(req frame.Request, compress, tracing bool) (frame.Res
 
 	streamID, err := c.r.setHandler(h)
 	if err != nil {
-		return nil, fmt.Errorf("setHandler: %w", err)
+		return nil, fmt.Errorf("set handler: %w", err)
 	}
 
 	r := request{
--- a/transport/conn.go
+++ b/transport/conn.go
@@ -89,18 +89,21 @@ type connReader struct {
 	buf  frame.Buffer
 	bufw io.Writer
 
-	h  map[frame.StreamID]responseHandler
+	h map[frame.StreamID]responseHandler
+	s streamIDAllocator
+	// mu guards h and s.
 	mu sync.Mutex
-
-	stream StreamIDAllocator
 }
 
 func (c *connReader) setHandler(h responseHandler) (frame.StreamID, error) {
 	c.mu.Lock()
-	streamID, err := c.stream.Alloc()
-	if err == nil {
-		c.h[streamID] = h
+	streamID, err := c.s.Alloc()
+	if err != nil {
+		c.mu.Unlock()
+		return 0, fmt.Errorf("stream ID alloc: %w", err)
 	}
+
+	c.h[streamID] = h
 	c.mu.Unlock()
 	return streamID, err
 }
@@ -181,7 +184,7 @@ const (
 	ioBufferSize    = 8192
 )
 
-func WrapConn(conn net.Conn, s StreamIDAllocator) *Conn {
+func WrapConn(conn net.Conn) *Conn {
 	c := &Conn{
 		conn: conn,
 		w: connWriter{
@@ -189,9 +192,9 @@ func WrapConn(conn net.Conn, s StreamIDAllocator) *Conn {
 			requestCh: make(chan request, requestChanSize),
 		},
 		r: connReader{
-			conn:   bufio.NewReaderSize(conn, ioBufferSize),
-			h:      make(map[frame.StreamID]responseHandler),
-			stream: s,
+			conn: bufio.NewReaderSize(conn, ioBufferSize),
+			h:    make(map[frame.StreamID]responseHandler),
+			s:    streamIDAllocator{},
 		},
 	}
 	go c.w.loop()
@@ -211,7 +214,7 @@ func (c *Conn) sendRequest(req frame.Request, compress, tracing bool) (frame.Res
 
 	streamID, err := c.r.setHandler(h)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("setHandler: %w", err)
 	}
 
 	r := request{
@@ -225,7 +228,7 @@ func (c *Conn) sendRequest(req frame.Request, compress, tracing bool) (frame.Res
 	c.w.submit(r)
 
 	resp := <-h
-	c.r.stream.Free(streamID)
+	c.r.s.Free(streamID)
 
 	return resp.Response, resp.Err
 }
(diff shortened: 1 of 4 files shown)


transport/conn.go
```diff
@@ -91,12 +91,18 @@ type connReader struct {
 
 	h  map[frame.StreamID]responseHandler
 	mu sync.Mutex
+
```
mmatczuk, line 94: Let's move stream up to group 2, you could actually rename it to `s`.
It should be noted that mu protects both `h` and `s`.

transport/stream.go
```diff
@@ -1,10 +1,40 @@
...
 
 type StreamIDAllocator interface {
 	Alloc() (frame.StreamID, error)
 	Free(id frame.StreamID)
 }
+
+const maxStreamID = math.MaxInt16
+
+const bucketSize = 64
+const buckets = (maxStreamID + 1) / bucketSize
```
mmatczuk, line 18: Please group consts

transport/stream.go
```diff
@@ -1,10 +1,40 @@
...
+
+// DefaultStreamIDAllocator is a StreamIDAllocator that always allocates the smallest possible stream on Alloc().
+type DefaultStreamIDAllocator struct {
+	usedBitmap [buckets]uint64
+}
+
+func (s *DefaultStreamIDAllocator) Alloc() (frame.StreamID, error) {
```
mmatczuk, line 25: Since this only holds a map you can pass it by value always.

transport/stream.go
```diff
@@ -1,10 +1,40 @@
...
+	for blockID, block := range &s.usedBitmap {
```
mmatczuk, line 26: Why address &?

transport/stream.go
```diff
@@ -1,10 +1,40 @@
...
+		if block < math.MaxUint64 {
+			offset := bits.TrailingZeros64(^block)
+			s.usedBitmap[blockID] |= 1 << offset
+			return frame.StreamID(offset + blockID*bucketSize), nil
+		}
+	}
+	return 0, fmt.Errorf("stream ID alloc: all stream ID's are busy")
```
mmatczuk, line 33:
```suggestion
	return 0, fmt.Error("all stream are busy")
```

mmatczuk: `stream ID alloc:` shall be added on caller level

mmatczuk (commented):
Please find comments inline.

Some followup items.
- We can remove the random stream ID generator
- We can remove the StreamIDAllocator interface 
- We can renam DefaultStreamIDAllocator to streamIDAllocator (not exported)
- We can then instantiate it in WrapConn since it is NOT shared

transport/stream.go
Kulezi: Correct me if I'm wrong but i think that the pointer receiver is needed as both Alloc() and Free() need to modify the map.

transport/stream.go
mmatczuk: This is because you are using an array! If you were using a slice it would be fine.

mmatczuk: The linter is correct.

transport/stream.go
mmatczuk: A) this is not a map - please keep the pointers - my bad.
B) map itself is kept on heap and what you have as map[X]Y is a pointer already, this is not like in c++

transport/conn.go
```diff
@@ -187,8 +194,8 @@ func WrapConn(conn net.Conn, s StreamIDAllocator) *Conn {
 		r: connReader{
 			conn: bufio.NewReaderSize(conn, ioBufferSize),
 			h:    make(map[frame.StreamID]responseHandler),
+			s:    streamIDAllocator{},
```
mmatczuk, line 197: The explicit struct initialization is not needed.

transport/conn_integration_test.go
```diff
@@ -14,7 +14,7 @@ func TestConnStartup(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	conn := WrapConn(nc, TestStreamIDAllocator{})
+	conn := WrapConn(nc, &DefaultStreamIDAllocator{})
```
mmatczuk, line 17: Likely it does not work

transport/conn.go
```diff
@@ -203,24 +210,25 @@ func (c *Conn) Startup(options frame.StartupOptions) (frame.Response, error) {
 }
 
 func (c *Conn) sendRequest(req frame.Request, compress, tracing bool) (frame.Response, error) {
-	streamID, err := c.stream.Alloc()
+	h := make(responseHandler)
+
+	streamID, err := c.r.setHandler(h)
 	if err != nil {
-		return nil, fmt.Errorf("stream ID alloc: %w", err)
+		return nil, fmt.Errorf("setHandler: %w", err)
```
mmatczuk, line 217: maybe `set handler` would be better.

transport/stream.go
```diff
@@ -1,10 +1,37 @@
...
+// Alloc() allocates the smallest possible stream ID on Alloc().
+func (s *streamIDAllocator) Alloc() (frame.StreamID, error) {
+	for blockID, block := range &s.usedBitmap {
+		if block < math.MaxUint64 {
+			offset := bits.TrailingZeros64(^block)
+			s.usedBitmap[blockID] |= 1 << offset
+			return frame.StreamID(offset + blockID*bucketSize), nil
+		}
+	}
+	return 0, fmt.Errorf("all stream ID's are busy")
```
mmatczuk, line 30: IDs

transport/stream_test.go
```diff
@@ -1,20 +1,67 @@
...
 import (
-	"math/rand"
-
 	"scylla-go-driver/frame"
+	"testing"
 )
 
-// TestStreamIDAllocator is StreamIDAllocator that returns a random.
-type TestStreamIDAllocator struct{}
+func TestDefaultStreamAllocatorAlloc(t *testing.T) {
```
mmatczuk, line 8: TestStreamAllocatorAlloc

transport/stream_test.go
```diff
@@ -1,20 +1,67 @@
...
-	const mask = int32(0xFFFF)
-	return frame.StreamID(rand.Int31() & mask), nil
+	// All streams are taken, we shouldn't be able to Alloc() another.
+	if _, err := s.Alloc(); err == nil {
+		t.Fatalf("allocating more than maxStreamID + 1 times in a row should fail")
+	}
 }
 
-func (t TestStreamIDAllocator) Free(id frame.StreamID) {
+func TestDefaultStreamAllocatorFree(t *testing.T) {
```
mmatczuk, line 38: It seems you could have a single test here. You can split them to different phases with subsets if you feel like separating them a little.

PR #113, opened 2022-01-24
Ml/dialing and opening connections

### LINKED ISSUE #104: Dialing / opening connections
- Add transport/routing.go with [ShardInfo](https://github.com/scylladb/scylla-rust-driver/blob/ee16c710e6a7164550f43a13c7125dbe644326cd/scylla/src/routing.rs#L25)
- Add transport.ConnectionConfig for the 1st version we need
    * pub tcp_nodelay: bool,
    * pub connect_timeout: std::time::Duration,
    * pub default_consistency: Consistency,
- Add transport.[OpenConnection](https://github.com/scylladb/scylla-rust-driver/blob/ee16c710e6a7164550f43a13c7125dbe644326cd/scylla/src/transport/connection.rs#L1125) note that the OpenConnection function family should only do the dialing and Wrap the TCP connection. The protocol handshake shall be implemented as an init method called by `WrapConn` function.
- We should only support shard aware ports i.e. dialing with source port.

Michal-Leszczynski:
Transport: conn, created family of three OpenConn functions able to establish connection.
Transport: routing, added simple functions for randomly choosing specific kind of local port (Rust approach).

Fixes #104

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -6,3 +6,4 @@ services:
     command: ${SCYLLA_ARGS}
     ports:
       - "9042:9042"
+      - "19042:19042"
--- a/transport/conn.go
+++ b/transport/conn.go
@@ -8,7 +8,9 @@ import (
 	"log"
 	"net"
 	"runtime"
+	"strconv"
 	"sync"
+	"time"
 
 	"scylla-go-driver/frame"
 	. "scylla-go-driver/frame/request"
@@ -167,6 +169,8 @@ func (c *connReader) parse(op frame.OpCode) frame.Response {
 		return ParseError(&c.buf)
 	case frame.OpReady:
 		return ParseReady(&c.buf)
+	case frame.OpSupported:
+		return ParseSupported(&c.buf)
 	default:
 		log.Fatalf("not supported %d", op)
 		return nil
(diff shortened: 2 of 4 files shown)


mmatczuk:
Cheers for detailed commit message, can you an empty line between header and body please.

transport/conn.go
```diff
@@ -172,11 +174,61 @@ type Conn struct {
 	stream StreamIDAllocator
 }
 
+type ConnConfig struct {
+	tcpNoDelay bool
+	timeout    time.Duration
```
mmatczuk, line 179: This must be exported, start with T not t

transport/conn.go
```diff
@@ -172,11 +174,61 @@ type Conn struct {
...
+	// defaultConsistency frame.Consistency
+}
+
 const (
 	requestChanSize = 1024
 	ioBufferSize    = 8192
 )
 
+// OpenShardConn opens connection mapped to a specific shard on scylla node.
+func OpenShardConn(rAddr string, si ShardInfo, cfg ConnConfig) (*Conn, error) { // nolint:unused // This will be used.
```
mmatczuk, line 190: I'd suggest and naming scheme change

- `rAddr` to `addr`
- `lAddr` to `localAddr`

transport/routing.go
```diff
@@ -0,0 +1,39 @@
...
+
+const (
+	// Range of ports that can be used to establish connection.
+	maxPort = 65535
+	minPort = 49152
+)
+
+type ShardInfo struct {
+	shard    uint16
+	nrShards uint16
```
mmatczuk, line 13: Please export fields of exported structs.

transport/routing.go
```diff
@@ -0,0 +1,39 @@
...
+	return r/si.nrShards*si.nrShards + si.shard
+}
+
+// ShardPortIterator returns iterator for consecutive ports that are
+// mapped to a specific shard on scylla node.
+func ShardPortIterator(si ShardInfo) func() uint16 {
+	curr := RandomShardPort(si)
+
+	return func() uint16 {
+		curr += si.nrShards
```
mmatczuk, line 33: cur?

transport/conn.go
```diff
@@ -172,11 +174,61 @@ type Conn struct {
...
+	conn, err := d.Dial("tcp", rAddr)
+	if err != nil {
+		return nil, fmt.Errorf("dialing TCP address %s: %w", rAddr, err)
+	}
+
+	tcpConn := conn.(*net.TCPConn)
+	if err = tcpConn.SetNoDelay(cfg.tcpNoDelay); err != nil {
+		return nil, fmt.Errorf("setting TCP no delay option: %w", err)
+	}
+	// TODO: add StreamIDAllocator.
```
mmatczuk, line 228: This will be removed.

transport/routing.go
Michal-Leszczynski: It was supposed to mean current. Will currentPort be more descriptive?

docker-compose.yml
```diff
@@ -6,3 +6,4 @@ services:
     command: ${SCYLLA_ARGS}
     ports:
       - "9042:9042"
+      - "19042:19042"
```
mmatczuk, line 9: 👍

transport/routing.go
```diff
@@ -0,0 +1,39 @@
...
+	maxRange := int(maxPort - si.NrShards + 1)
+	minRange := int(minPort + si.NrShards - 1)
+	r := uint16(rand.Intn(maxRange-minRange+1) + minRange)
+	return r/si.NrShards*si.NrShards + si.Shard
+}
+
+// ShardPortIterator returns iterator for consecutive ports that are
+// mapped to a specific shard on scylla node.
+func ShardPortIterator(si ShardInfo) func() uint16 {
+	currentPort := RandomShardPort(si)
```
mmatczuk, line 30: cur was fine port would be also fine.

Follow the rule of thumb that the length of var name represents its scope.
Here scope is small but not minuscule :semi

PR #117, opened 2022-01-28
various improvements and extinctions to code

mmatczuk:
Notable changes:

* Added Makefile targets to operate docker-compose Scylla instance
* Added Makefile targets to run integration tests and benchmarks
* Added .gitignore


mmatczuk:
PTAL

mmatczuk:
@Michal-Leszczynski for dialing to work you can use the "host" network in docker.
This unfortunately works under Linux only we'd have to fix that anyways.

.gitignore
```diff
@@ -1 +1,19 @@
-.idea/
\ No newline at end of file
+./.idea
```
Choraden, line 1: I'm not sure if this is intentional, but in my case .idea is placed in the same dir as .git
please add .idea

PR #112, opened 2022-01-24
Node connection pool

Neofine:

PR #119, opened 2022-02-09
Ml/close

Michal-Leszczynski:
Added conn.Close() for closing connections.

### DIFF
--- a/transport/conn.go
+++ b/transport/conn.go
@@ -142,15 +142,6 @@ func (c *connReader) handler(streamID frame.StreamID) responseHandler {
 	return h
 }
 
-func (c *connReader) drainHandlers() {
-	c.mu.Lock()
-	c.closed = true
-	for _, h := range c.h {
-		h <- response{Err: fmt.Errorf("connection closed")}
-	}
-	c.mu.Unlock()
-}
-
 func (c *connReader) loop() {
 	runtime.LockOSThread()
 
@@ -166,10 +157,20 @@ func (c *connReader) loop() {
 		if h := c.handler(resp.StreamID); h != nil {
 			h <- resp
 		} else {
+			// Connection is corrupted, must be closed.
 			c.drainHandlers()
 			return
 		}
 	}
+}
+
+func (c *connReader) drainHandlers() {
+	c.mu.Lock()
+	c.closed = true
+	for _, h := range c.h {
+		h <- response{Err: fmt.Errorf("connection closed")}
+	}
+	c.mu.Unlock()
 }
 
 func (c *connReader) recv() response {
@@ -235,10 +236,6 @@ type ConnConfig struct {
 const (
 	requestChanSize = 1024
 	ioBufferSize    = 8192
-
-	// Each handler may encounter 2 responses, one from connWriter.loop()
-	// and one from drainHandlers().
-	responseHandlerCapacity = 2
 )
 
 // OpenShardConn opens connection mapped to a specific shard on scylla node.
@@ -350,7 +347,9 @@ func (c *Conn) Query(s Statement, pagingState frame.Bytes) (QueryResult, error) 
 }
 
 func (c *Conn) sendRequest(req frame.Request, compress, tracing bool) (frame.Response, error) {
-	h := make(responseHandler, responseHandlerCapacity)
+	// Each handler may encounter 2 responses, one from connWriter.loop()
+	// and one from drainHandlers().
+	h := make(responseHandler, 2)
 
 	streamID, err := c.r.setHandler(h)
 	if err != nil {
@@ -365,7 +364,7 @@ func (c *Conn) sendRequest(req frame.Request, compress, tracing bool) (frame.Res
 		ResponseHandler: h,
 	}
 
-	// Note: requestCh might be full after terminating writeLoop so some goroutines could hang here forever.
+	// requestCh might be full after terminating writeLoop so some goroutines could hang here forever.
 	// this could be fixed by changing requestChanSize to be able to hold all possible streamIDs,
 	// adding a grace period before terminating writeLoop or counting active streams.
 	c.w.submit(r)
--- a/transport/conn.go
+++ b/transport/conn.go
@@ -9,7 +9,6 @@ import (
 	"net"
 	"runtime"
 	"strconv"
-	"strings"
 	"sync"
 	"time"
 
@@ -144,18 +143,11 @@ func (c *connReader) loop() {
 	for {
 		resp := c.recv()
 
-		// Checks if connection was closed.
-		if resp.Err != nil && strings.Contains(resp.Err.Error(), net.ErrClosed.Error()) {
-			// Not sure about this checking, maybe it would be easier
-			// to put unchanged error inside resp in recv.
-			return
-		}
-
 		if h := c.handler(resp.StreamID); h != nil {
 			h <- resp
 		} else {
-			// FIXME gracefully handle recv error
-			log.Fatalf("recv error: %+v, %+v, %v", resp.Header, resp.Response, resp.Err)
+			// Connection is corrupted. Must be closed.
+			return
 		}
 	}
 }
@@ -205,6 +197,14 @@ func (c *connReader) parse(op frame.OpCode) frame.Response {
 		log.Fatalf("not supported %d", op)
 		return nil
 	}
+}
+
+func (c *connReader) clearHandlers() {
+	c.mu.Lock()
+	for _, h := range c.h {
+		h <- response{Err: fmt.Errorf("connection closed")}
+	}
+	c.mu.Unlock()
 }
 
 type Conn struct {
@@ -314,9 +314,10 @@ func (c *Conn) init() error {
 }
 
 // Close closes connection and terminates reader and writer go rutines.
-func (c *Conn) Close() error {
+func (c *Conn) Close() {
+	_ = c.conn.Close()
 	close(c.w.requestCh)
-	return c.conn.Close()
+	c.r.clearHandlers()
 }
 
 func (c *Conn) Startup(options frame.StartupOptions) (frame.Response, error) {
--- a/transport/conn.go
+++ b/transport/conn.go
@@ -9,6 +9,7 @@ import (
 	"net"
 	"runtime"
 	"strconv"
+	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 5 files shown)


transport/conn_integration_test.go
```diff
@@ -3,42 +3,34 @@
...
 
-func TestConnStartup(t *testing.T) {
-	//nc, err := net.Dial("tcp", "localhost:9042")
-	//if err != nil {
-	//	t.Fatal(err)
-	//}
-	//conn := WrapConn(nc, TestStreamIDAllocator{})
+func TestMain(m *testing.M) {
+	goleak.VerifyTestMain(m)
+}
```
mmatczuk, line 17: extract to main_test.go

transport/conn.go
```diff
@@ -130,11 +127,12 @@ func (c *connReader) loop() {
 	c.bufw = frame.BufferWriter(&c.buf)
 	for {
 		resp := c.recv()
+
```
mmatczuk, line 130: ?

transport/conn.go
```diff
@@ -289,7 +295,12 @@ func (c *Conn) init() error {
 	}
 }
 
-// TODO add conn Close, make sure go routines exit
+// Close closes connection and terminates reader and writer go rutines.
+func (c *Conn) Close() {
```
mmatczuk, line 299: Add closed bool protected by mu.

mmatczuk: If closed send should report error immediately.

transport/conn.go
```diff
@@ -289,7 +295,12 @@ func (c *Conn) init() error {
...
+	_ = c.conn.Close()
+	close(c.w.requestCh)
+	c.r.clearHandlers()
```
mmatczuk, line 302: This should be called drainHandlers, close should be added to clearHandlers renamed to drainHandlers

transport/conn.go
```diff
@@ -186,6 +184,14 @@ func (c *connReader) parse(op frame.OpCode) frame.Response {
 	}
 }
 
+func (c *connReader) clearHandlers() {
+	c.mu.Lock()
```
mmatczuk, line 188: drainHanldersLocked - no need to take a lock here, Close shall own the locking.

transport/conn.go
```diff
@@ -106,19 +106,16 @@ type connReader struct {
 	h map[frame.StreamID]responseHandler
 	s streamIDAllocator
 	// mu guards h and s.
-	mu sync.Mutex
+	mu *sync.Mutex
```
mmatczuk, line 109: This is not needed, address is taken relative to connReader ptr.

transport/conn.go
```diff
@@ -323,8 +342,15 @@ func (c *Conn) Query(s Statement, pagingState frame.Bytes) (QueryResult, error)
 func (c *Conn) sendRequest(req frame.Request, compress, tracing bool) (frame.Response, error) {
 	h := make(responseHandler)
 
-	streamID, err := c.r.setHandler(h)
+	c.mu.Lock()
```
mmatczuk, line 345: Conn shall not manage the connReader maybe the mutex should be migrated to conn.

transport/conn.go
```diff
@@ -304,7 +313,17 @@ func (c *Conn) init() error {
...
 }
 
-// TODO add conn Close, make sure go routines exit
+// Close closes connection and terminates reader and writer go routines.
+func (c *Conn) Close() {
+	c.mu.Lock()
+	c.closed = true
+
+	_ = c.conn.Close()
+	close(c.w.requestCh)
```
mmatczuk, line 322: Move to drain

transport/conn.go
```diff
@@ -304,7 +313,17 @@ func (c *Conn) init() error {
...
+	c.r.drainHandlersLocked()
+
+	c.mu.Unlock()
```
mmatczuk, line 325: use defer

mmatczuk:
Please revert the last commit.

transport/conn.go
```diff
@@ -279,6 +286,8 @@ func WrapConn(conn net.Conn) (*Conn, error) {
 			h:    make(map[frame.StreamID]responseHandler),
 		},
 	}
+	c.r.mu = &c.mu
```
mmatczuk, line 289: Mutex sharing, hmmm

Choraden:
After thinking of it once again, I'll propose different approach. 
We have to take into consideration that closing connection is in relation with connection pool and refiller.
 
Let's assume ConnPool will have get method, that returns connection, which will be used **instantly** and **only once**. 
```go 
ConnPool.get(..) -> *Conn
```
I understand that the point of having mutex here is that sending on closed channel will panic.
But what if 
```go
connection.Close() {
        // Close TCP socket.
        // Terminate reading loop and respond to remaining request with error (drain streamId map).
	connection.conn.Close() 
	go func() {
	        // Give some time for possible requests to be send via request channel.
                // Socket is closed so any request would be cancelled anyway. 
		time.Sleep(5 * time.Second) 
		close(requestCh) 
               // Once channel is closed writing loop will drain requestCh and respond with error the each request,
               // so they eventaully will be retried. 
               // Writing loop is terminated due to empty and closed requestCh
	}()	
}
```

Assuming that connection failures are rare, I think we would benefit from creating new goroutine to close connection, instead of taking mutex each time we send request.

This way we could also omit mutex in connectionPool as we are now not afraid of using broken connections, which firstly will be removed from the pool and followingly closed.  

Please ask if something is unclear and let me know what you guys think

Choraden:
^ This may have influence on #112 @Neofine

mmatczuk:
@Choraden adding a grace period like that can be added - it affects the pool impl more than the actual closing.

For close the requirements are:
* The underlying TCP conn is closed
* The go routines are stopped
* All in-flight requests are errors
* If adding new send request if fails immediately 
* Code does not panic

Kulezi:
Here is an approach that is correct, graceful and etc. but uses a mutex, we need one mutex anyway because streamID allocation needs atomicity, so it might not cost that much. 
<details>
<summary> Mutex approach  summary </summary>
In this approach:

   - We will never try to write to closed requestCh because
    allocating streamID for request will return error
    when any side of connection is closed.
   - No double closing happens.
   - All handlers are drained, as after draining all calls to allocation return error
   - Writer loop stops because it can detect a closed channel.
   - Read loop stops because recv on closed connection will return an error.

conn.mu refers to the whole connection - that's the reason there was mutex sharing in last commit
```
sendRequest():
    lock conn.mu
    allocate streamID for request and
    send request over requestCh
    unlock conn.mu
    read response from handler
    lock conn.mu
    free streamID for request (note: it may have been already drained so check for that)
    unlock conn.mu 

connReader.loop():
    for {
        read response
        on error that should close the connection
            lock conn.mu
            if conn.readClosed is false
                set conn.readClosed to true
                drain handlers
                notify pool that connection needs to be closed
            unlock conn.mu
            exit the loop

        lock conn.mu
        fetch handler for request
        send response to handler
        unlock conn.mu
    }

connWriter.loop():
    get request from requestCh
    try sending it
    on error that should close the connection
        lock conn.mu
        if conn.writeClosed is false
            set conn.writeClosed to true 
            close(requestCh)
            notify pool that connection needs to be closed
        unlock conn.mu
        exit the loop

conn.Close(): <- this should be called by the connection owner (pool)
    lock conn.mu
    if conn.readClosed is false
        set conn.readClosed to true
        drain handlers

    if conn.writeClosed is false
        set conn.writeClosed to true
        close(requestCh)

    unlock conn.mu
```
</details>


@Choraden I think that draining before the sleep can make the writerLoop wait forever, consider following scenario:
request is sent through requestCh
Close() happens, streams are drained, request above gets a connection closed
however the request in requestCh is still there
after the wait we pick it in writerLoop and try to send a error to the handler
but no one is on the reading side of handler and we get blocked.
So we either have to make the channel buffered or move the draining somewhere else
Im not yet sure what happens when we need to close connection from reader/writer loops end but it should be doable 

I think that your approach would simplify the code, if connections don't close often it probably would be a bit of a performance booster. 

@mmatczuk let me know which approach we should choose

mmatczuk:
My notes on implementation:

1. It the `requestCh` is closed the `submit` func shall call `defer recover()` - hence it's better to never close it
2. `ResponseHandler` needs to become a buffered channel with buffering of 1 (reader or writer) otherwise go routines can block preventing any draining...
3. Close should only close the underlying connection, go routines should pick up the I/O error drain and exit
  * connWriter drains the channel, no error is reported, under the assumption that you can only submit sth that has frame ID assigned ergo has an entry in the handlerMap
  * connReader takes a lock and drains handler map (reports error to ResponseHandler), and sets map to nil
4. connReader may add closed bool flag for improved readability, handler map is nil iff closed is set to true
5. connReader setHandler and freeHandler respect the closed flag

mmatczuk:
@Kulezi 

Sharing mutex pointer and/or managing connReader mutex from Conn is a no go from code maintenance pov.

Submit under a mutex is also not great but honestly the mutex only makes sense in connReader and extending it to Conn has no real justification.

Closing the channel (if not under mutex) may result panic.

There is no error handling in connReader and not sufficient error handling in connWriter.

It requires more testing as getting it wrong can result in go routine leaks, queries being stalled and panics...

Choraden:
Writer loop might be blocked on requestCh recv, therefore it must be notified to end its work.

transport/conn.go
```diff
@@ -52,21 +54,20 @@ func (c *connWriter) loop() {
 	var maxCoalescedRequests = requestChanSize / 10
 
 	for {
-		r, ok := <-c.requestCh
-		if !ok {
-			return
-		}
-
-		size := len(c.requestCh) + 1
+		size := len(c.requestCh)
```
mmatczuk, line 57: 👍

transport/conn.go
```diff
@@ -136,17 +142,32 @@ func (c *connReader) handler(streamID frame.StreamID) responseHandler {
...
+
+		if resp.Err != nil {
+			c.drainHandlers()
+			return
+		}
+
 		if h := c.handler(resp.StreamID); h != nil {
 			h <- resp
 		} else {
-			// FIXME gracefully handle recv error
```
mmatczuk, line 148: Would be good to add a comment here.

transport/conn.go
```diff
@@ -136,17 +142,32 @@ func (c *connReader) handler(streamID frame.StreamID) responseHandler {
 	return h
 }
 
+func (c *connReader) drainHandlers() {
```
mmatczuk, line 145: Can we move it below loop?

transport/conn.go
```diff
@@ -214,6 +235,10 @@ type ConnConfig struct {
 const (
 	requestChanSize = 1024
 	ioBufferSize    = 8192
+
+	// Each handler may encounter 2 responses, one from connWriter.loop()
+	// and one from drainHandlers().
+	responseHandlerCapacity = 2
```
mmatczuk, line 241: I'd suggest to inline that and move the comment.
The comment here is lacking context and the value is not a subject to configuration.

transport/conn.go
```diff
@@ -336,6 +365,9 @@ func (c *Conn) sendRequest(req frame.Request, compress, tracing bool) (frame.Res
 		ResponseHandler: h,
 	}
 
+	// Note: requestCh might be full after terminating writeLoop so some goroutines could hang here forever.
```
mmatczuk, line 368: Note is not needed just write what you have to say.
This is a feature, it could be put as a doc string of sendRequest indicating that it can block.

mmatczuk:
Overall it looks good, left some inline comments.

transport/conn.go
Kulezi: I don't think that sendRequest possibly hanging forever is a good feature. Currently we are not guaranteed that closeRequest will be the last thing ever sent through requestCh, if more than requestChanSize requests will pile up after it those goroutines will get stuck forever.

PR #125, opened 2022-02-20
Frame: Added ScyllaSupported

Neofine:
ScyllaSupported represents Scylla connection options as sent in SUPPORTED frame.
It will be used in f.e. PoolRefiller from issue #112.

### DIFF
--- a/frame/types.go
+++ b/frame/types.go
@@ -28,6 +28,29 @@ type Inet struct {
 	Port Int
 }
 
+// ScyllaSupported represents Scylla connection options as sent in SUPPORTED frame.
+// https://github.com/scylladb/scylla/blob/master/docs/design-notes/protocol-extensions.md#intranode-sharding
+type ScyllaSupported struct {
+	Shard             uint16
+	NrShards          uint16
+	MsbIgnore         uint8
+	Partitioner       string
+	ShardingAlgorithm string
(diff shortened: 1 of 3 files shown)


frame/types.go
```diff
@@ -28,6 +28,29 @@ type Inet struct {
...
+// https://github.com/scylladb/scylla/blob/master/docs/design-notes/protocol-extensions.md#intranode-sharding
+type ScyllaSupported struct {
+	Shard             uint16
+	NrShards          uint16
+	MsbIgnore         uint8
+	Partitioner       string
+	ShardingAlgorithm string
+	ShardAwarePort    uint16
+	ShardAwarePortSSL uint16
+	LwtFlagMask       int
```
Neofine, line 41: 'LwtFlagMask' is present in gocql implementation but is nowhere to be used and does not appear to have its own flag in SUPPORTED, the only mention about it is in
https://github.com/scylladb/scylla/blob/master/docs/design-notes/protocol-extensions.md#lwt-prepared-statements-metadata-mark
where there are two additional flags for it, should we jump into this rabbit hole and try to implement this, just delete this flag from ScyllaSupported or keep it, what do you think @mmatczuk ?

mmatczuk: Leave it for now.

PR #128, opened 2022-02-23
Connection Pool and Pool Refiller

Neofine:


Neofine:
When reviewing this PR please limit yourselves to the latest commit, first 5 commits have been made regarding #119 and probably should be reviewed there.

Neofine:
```go
// RandConn returns a random connection but has no protection
// of it being nil, that's why NextConn() might be a better choice for now.
func (c *ConnPool) RandConn() *Conn {
	idx := rand.Uint32() % uint32(c.nrShards)
	return c.loadConn(uint16(idx))
}

// NextConn returns nearest (from idx % nrShards) not nil connection
// if it iterates over whole slice and doesn't find any (extremely unlikely) it returns nil.
func (c *ConnPool) NextConn() *Conn {
	idx := c.idx.Add(1)
	for i := uint16(0); i < c.nrShards; i++ {
		if conn := c.loadConn(uint16((idx + uint64(i)) % uint64(c.nrShards))); conn != nil {
			return conn
		}
	}
	return nil
}
```

If this doesn't hurt our performance nor correctness I think second approach of getting connection from ConnPool is better as it isn't as error prone as first one. Second one is inspired by https://github.com/scylladb/gocql/blob/master/scylla.go#L303

Neofine:
In this approach ```go func (c *ConnPool) Close(conn *Conn)  ``` is used only to send a message to the PoolRefiller because it would save us some fuss regarding multithreading and not using mutexes, as there is only one instance of PoolRefiller it will close not disturbed those connections without any need for more atomicity in variables.

Will this approach hurt our performance? Should it be benchmarked @mmatczuk ?

transport/main_test.go
```diff
@@ -1,11 +1,6 @@
...
-import (
-	"testing"
-
-	"go.uber.org/goleak"
-)
-
-func TestMain(m *testing.M) {
-	goleak.VerifyTestMain(m)
-}
+// Commented out, PoolRefiller at this time doesn't have an option to close its loop().
```
mmatczuk, line 3: Please rebase and enable that.

transport/pool.go
```diff
@@ -0,0 +1,252 @@
...
+// Node Connection Pool - One per node, Node struct holds pointer to it
+// This struct holds connection to shards along with helper variables.
+type ConnPool struct {
+	conns     []atomic.Value
+	alertChan chan uint16
+
+	nrShards  uint16
+	msbIgnore uint8
+
+	idx atomic.Uint64
```
mmatczuk, line 26: Grouping is not needed and conns should be close to idx imo.
Can we have it like G1,G0,G2?

transport/pool.go
```diff
@@ -0,0 +1,252 @@
...
+}
+
+// Atomicly stores pointer to a connection.
+func (c *ConnPool) storeConn(idx uint16, conn *Conn) {
+	c.conns[idx].Store(conn)
+}
+
+// Atomicly load pointer to a connection.
+func (c *ConnPool) loadConn(idx uint16) *Conn {
+	conn := c.conns[idx].Load()
```
mmatczuk, line 36: You could cast it like that `conn, _ := c.conns[idx].Load().(*Conn)`

transport/pool.go
```diff
@@ -0,0 +1,252 @@
...
+
+// RandConn returns a random connection but has no protection
+// of it being nil, that's why NextConn() might be a better choice for now.
+func (c *ConnPool) RandConn() *Conn {
+	idx := rand.Uint32() % uint32(c.nrShards)
+	return c.loadConn(uint16(idx))
+}
+
+// NextConn returns nearest (from idx % nrShards) not nil connection
+// if it iterates over whole slice and doesn't find any (extremely unlikely) it returns nil.
```
mmatczuk, line 51: No need to describe implementation, it's effectively plain round robin, can return nil if pool is empty.

transport/pool.go
```diff
@@ -0,0 +1,252 @@
...
+	idx := c.idx.Add(1)
+	for i := uint16(0); i < c.nrShards; i++ {
+		if conn := c.loadConn(uint16((idx + uint64(i)) % uint64(c.nrShards))); conn != nil {
+			return conn
+		}
+	}
+	return nil
+}
+
+// Calculates shard index in ConnPool.conns slice based on given token
```
mmatczuk, line 62: ConnPool.conns is private no need to mention it.
Dot at the end.

transport/pool.go
```diff
@@ -0,0 +1,252 @@
...
+	. "scylla-go-driver/frame/response"
+
+	"go.uber.org/atomic"
+)
+
+// Node Connection Pool - One per node, Node struct holds pointer to it
+// This struct holds connection to shards along with helper variables.
+type ConnPool struct {
+	conns     []atomic.Value
+	alertChan chan uint16
```
mmatczuk, line 21: Can we have closeCh, a channel to close a connection under an index.

mmatczuk:
As a general comment ConnPool shall never return null connection if it's not empty i.e. RandomConn should fallback to NextConn.

transport/pool.go
```diff
@@ -0,0 +1,252 @@
...
+		// so let's give it to the caller.
+		return conn
+	}
+	return c.RandConn()
+}
+
+// Helper in closing connection, actual closing happens
+// in PoolRefiller because of multithreading issues.
+// When wanted connection is found it passes its index to PoolRefiller channel.
+func (c *ConnPool) Close(conn *Conn) {
```
mmatczuk, line 90: Let's add shard nr to Conn.

mmatczuk:
Another general comments docs describe what a function / struct provides, or requires not what it does in terms of implementation. You can keep the implementation notes as comments if that's actually helpful but it's not useful in docs and can easily change.

transport/pool.go
```diff
@@ -0,0 +1,252 @@
...
+	supp, ok := res.(*Supported)
+	if !ok {
+		return frame.ScyllaSupported{}, errors.New("couldn't cast interface to struct")
+	}
+
+	return *supp.ParseScyllaSupported(), nil
+}
+
+// Initiates ConnPool along with its PoolRefiller.
+func InitNodeConnPool(addr string, connConfig ConnConfig) *ConnPool { // nolint:unused // This will be used.
```
mmatczuk, line 244: Docs on exported functions must start with the function name there is a linter for that.

mmatczuk: Better add a dummy use.

transport/pool.go
```diff
@@ -0,0 +1,252 @@
...
+// When wanted connection is found it passes its index to PoolRefiller channel.
+func (c *ConnPool) Close(conn *Conn) {
+	for i, v := range c.conns {
+		if v.Load() == conn {
+			c.alertChan <- uint16(i)
+		}
+	}
+}
+
+// Handles errors and occasionally fills ConnPool.
```
mmatczuk, line 98: occasionally?

mmatczuk: ?

transport/pool.go
```diff
@@ -0,0 +1,252 @@
...
+type PoolRefiller struct {
+	address    string
+	connConfig ConnConfig
+	connPool   ConnPool
+
+	activeShards uint16
+
+	connOptions      frame.ScyllaSupported
+	closeSinceRefill bool
```
mmatczuk, line 107: What's that?

transport/pool.go
```diff
@@ -0,0 +1,252 @@
...
+	currentDelay     time.Duration
```
mmatczuk, line 108: What's that?

transport/pool.go
```diff
@@ -0,0 +1,252 @@
...
+
+// Handles errors and occasionally fills ConnPool.
+type PoolRefiller struct {
+	address    string
+	connConfig ConnConfig
+	connPool   ConnPool
+
+	activeShards uint16
+
+	connOptions      frame.ScyllaSupported
```
mmatczuk, line 106: This should go above connPool?

transport/pool.go
```diff
@@ -0,0 +1,252 @@
...
+func (c *ConnPool) Close(conn *Conn) {
+	for i, v := range c.conns {
+		if v.Load() == conn {
+			c.alertChan <- uint16(i)
+		}
+	}
+}
+
+// Handles errors and occasionally fills ConnPool.
+type PoolRefiller struct {
```
mmatczuk, line 99: Grouping probably not needed.

transport/routing.go
```diff
@@ -13,8 +13,10 @@ const (
 )
 
 type ShardInfo struct {
-	Shard    uint16
-	NrShards uint16
+	Shard     uint16
+	NrShards  uint16
+	MsbIgnore uint8
+
 	// This will be used.
```
mmatczuk, line 20: We can remove it.

mmatczuk:
Please save time by not adding useless comments to private methods, example

```
// Atomicly stores pointer to a connection.
 func (c *ConnPool) storeConn(idx uint16, conn *Conn)
```

It's already clear that is stores a connection.

transport/pool.go
```diff
@@ -0,0 +1,252 @@
...
+	ticker := time.NewTicker(p.currentDelay)
+	refilling := true
+
+	for {
+		// Wakes up either when it's refilling time or when there is connection to be closed.
+		select {
+		case <-ticker.C:
+			p.fill()
+			refilling = false
+		case idx := <-p.connPool.alertChan:
```
mmatczuk, line 132: Let's use a function in p.connPool to get the channel, like in context.Done()

transport/pool.go
```diff
@@ -0,0 +1,252 @@
...
+	p.connPool.nrShards = p.connOptions.NrShards
+	p.connPool.conns = make([]atomic.Value, p.connPool.nrShards)
+	p.connPool.storeConn(p.connOptions.Shard, conn)
+	p.connPool.msbIgnore = p.connOptions.MsbIgnore
+	p.connPool.alertChan = make(chan uint16, p.connPool.nrShards)
+	p.activeShards = 1
+}
+
+// Refills ConnPool.
+func (p *PoolRefiller) fill() {
```
mmatczuk, line 202: Can we move that just before loop and put other functions below in the order of dependency / appearance.

The objective is to be able to read it top down without too much jumping.

transport/pool.go
```diff
@@ -0,0 +1,252 @@
...
+
+// Fill happens in changing time intervals, these are its helper constants.
+const (
+	minFillBackoff        = 50 * time.Millisecond
+	maxFillBackoff        = 10 * time.Second
+	fillBackoffMultiplier = 2
+)
+
+// PoolRefiller endless function, for now it never shuts down.
+func (p *PoolRefiller) loop() {
```
mmatczuk, line 119: This is a quite useless comment `PoolRefiller endless function, for now it never shuts down.`

BUT

every go loop function must be able to exit - think we close the session and.

transport/pool.go
```diff
@@ -0,0 +1,252 @@
...
+			// Time interval of refilling is adjusted based on what caused it.
+			if p.closeSinceRefill {
+				p.onClose()
+			} else {
+				p.currentDelay = minFillBackoff
+			}
+
+			ticker.Reset(p.currentDelay)
+			refilling = true
+		} else if !p.needsFilling() {
```
mmatczuk, line 146: Can we rewrite the if branching to improve readability.

transport/pool.go
```diff
@@ -0,0 +1,252 @@
...
+	p.connPool.conns = make([]atomic.Value, p.connPool.nrShards)
+	p.connPool.storeConn(p.connOptions.Shard, conn)
+	p.connPool.msbIgnore = p.connOptions.MsbIgnore
+	p.connPool.alertChan = make(chan uint16, p.connPool.nrShards)
+	p.activeShards = 1
+}
+
+// Refills ConnPool.
+func (p *PoolRefiller) fill() {
+	if p.activeShards == 0 && p.connPool.nrShards == 0 {
```
mmatczuk, line 203: maybe check conn pool size?

transport/pool.go
```diff
@@ -0,0 +1,252 @@
...
+			MsbIgnore: p.connPool.msbIgnore,
+		}
+
+		conn, err := OpenShardConn(p.address, shard, p.connConfig)
+		if err == nil {
+			p.connPool.storeConn(shard.Shard, conn)
+			p.activeShards++
+		} else {
+			p.closeSinceRefill = true
+		}
```
mmatczuk, line 225: Typically you handle err != nil and then call continue

transport/pool.go
```diff
@@ -0,0 +1,252 @@
...
+	}
+}
+
+// Sends request for frame.Supported struct and returns either frame.ScyllaSupported or error.
+func getScyllaSupported(conn *Conn) (frame.ScyllaSupported, error) {
```
mmatczuk, line 230: This shall be moved to Conn

mmatczuk:
Conn needs to be integrated with ConnPool / Refiller to replace connection on error.
Let's not use exponential backoff at the moment to simplify code.
Let's wake every 1s but try to fill immediately after close.
First connection needs to be established on ConnPool creation, the creation may result in an error.

mmatczuk:
There needs to be a Close method on ConnPool that closes all connections and halts refiller.

Neofine:
Should
 ```go func (c *ConnPool) RandConn() *Conn ``` 
and
 ```go func (c *ConnPool) NextConn() *Conn ``` 
be visible outside pool.go or just RandConn as it uses NextConn as backup?

transport/conn.go
```diff
@@ -221,10 +221,11 @@ func (c *connReader) parse(op frame.OpCode) frame.Response {
...
 type Conn struct {
-	conn  net.Conn
-	shard uint16
-	w     connWriter
-	r     connReader
+	conn    net.Conn
+	shard   uint16
+	w       connWriter
+	r       connReader
+	closeCh chan int32
```
mmatczuk, line 228: ?

mmatczuk:
Please do not add Merge commits @Neofine, the main commit contains shitloads of other work WTF?

transport/conn.go
```diff
@@ -240,8 +241,8 @@ const (
 )
 
 // OpenShardConn opens connection mapped to a specific shard on Scylla node.
-func OpenShardConn(addr string, si ShardInfo, cfg ConnConfig) (*Conn, error) { // nolint:unused // This will be used.
-	it := ShardPortIterator(si)
+func OpenShardConn(addr string, si *ShardInfo, cfg ConnConfig, closeCh chan int32) (*Conn, error) { // nolint:unused // This will be used.
```
mmatczuk, line 244: ?

transport/conn.go
```diff
@@ -346,6 +349,14 @@ func (c *Conn) Supported() (*Supported, error) {
 	return nil, responseAsError(res)
 }
 
+func (c *Conn) GetScyllaSupported() (*ScyllaSupported, error) {
```
mmatczuk, line 352: Useless.

mmatczuk (approved):

mmatczuk:
@Neofine please read that.

transport/pool.go
```diff
@@ -0,0 +1,201 @@
...
+		if err != nil {
+			log.Printf("failed to open shard conn: %s", err)
+			continue
+		}
+		if conn.Shard() != i {
+			log.Fatalf("opened conn to wrong shard: expected %d got %d", i, conn.Shard())
+		}
+		conn.setOnClose(r.onConnClose)
+		r.pool.storeConn(conn)
+		r.active++
```
Choraden, line 191: Correct me if I'm wrong, but shouldn't we check if there already exists connection to this shard and then store it in pool? Wouldn't it cause any trouble to just throw away the pointer of working connection without calling Close() on it?

mmatczuk: > Correct me if I'm wrong, but shouldn't we check if there already exists connection to this shard and then store it in pool?

We do, note the line 176.

> Wouldn't it cause any trouble to just throw away the pointer of working connection without calling Close() on it?

It happens only when the connection is closed, first closed then removed.
It could also happen when the connection is broken, not yet implemented.
In both cases a dysfunctional connection can be returned.
It would be handled and retried.

Choraden: We do check i-th connection, but what I meant was the following scenario:
1. r.pool.LoadConn(0) == nil 
2. OpenShardConn opens connection to wrong shard  - conn.Shard = 1 (I don't know how likely it is to happen)
3. But we already have working connection to shard 1 
4. r.pool.StoreConn stores newly opened connection and loses pointer to already working connection to shard 1

I'm afraid we are not able to gracefully close such lost connection.

mmatczuk: log.Fatalf exits the process, it's a FATAL error if your connections goes to a different shard.

PR #129, opened 2022-02-24
experiments: Atomic load benchmark - atomic value vs unsafe pointer

Choraden:
I've benchmarked atomics which we will use in cluster. 
The winner is unsafe pointer.

Benchmark result:
goos: darwin
goarch: amd64
pkg: scylla-go-driver/experiments
cpu: Intel(R) Core(TM) i7-7700HQ CPU @ 2.80GHz
BenchmarkLoadAV
BenchmarkLoadAV-8   	691782937	         1.614 ns/op
BenchmarkLoadUP
BenchmarkLoadUP-8   	1000000000	         0.4589 ns/op
PASS

### DIFF
--- a/experiments/atomicbenchmark_test.go
+++ b/experiments/atomicbenchmark_test.go
@@ -4,6 +4,8 @@ import (
 	"sync/atomic"
 	"testing"
 	"unsafe"
+
+	uber "go.uber.org/atomic"
 )
 
 type data = map[string]int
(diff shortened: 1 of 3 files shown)


mmatczuk:
Cheers for the benchmark, this is the right approach.
It also shows that it costs 1 NS so there is also room for convenience :)

mmatczuk:
For convenience you can include github.com/uber-go/atomic in the benchmark

Choraden:
Benchmark with uber: 
goos: darwin
goarch: amd64
pkg: scylla-go-driver/experiments
cpu: Intel(R) Core(TM) i7-7700HQ CPU @ 2.80GHz
BenchmarkLoadAV
BenchmarkLoadAV-8       	864799339	         1.216 ns/op
BenchmarkLoadUP
BenchmarkLoadUP-8       	1000000000	         0.4040 ns/op
BenchmarkLoadUberAV
BenchmarkLoadUberAV-8   	946084138	         1.198 ns/op
BenchmarkLoadUberUP
BenchmarkLoadUberUP-8   	1000000000	         0.5388 ns/op
PASS

New benchmark has shown uber package is as fast as sync.atomic. 
When we look into uber implementation we can see it wraps sync.atomic, so basically these things are the same.

mmatczuk:
Yes it is there no way around it the API is easier to work with.

PR #130, opened 2022-02-24
Ml/new control

Michal-Leszczynski:


transport/cluster.go
```diff
@@ -0,0 +1,268 @@
...
+		Consistency: frame.ONE,
+	}
+
+	localQuery = Statement{
+		Content:     "SELECT rpc_address, data_center, rack, tokens FROM system.local",
+		Consistency: frame.ONE,
+	}
+)
+
+func makePeerStatement(addr frame.Inet) Statement {
```
mmatczuk, line 27: How about peersQuery?

transport/cluster.go
```diff
@@ -0,0 +1,268 @@
...
+	return Statement{
+		Content:     "SELECT peer, data_center, rack, tokens FROM system.peers WHERE peer = " + net.IP(addr.IP).String(),
+		Consistency: frame.ONE,
+	}
+}
+
+func InetToString(i frame.Inet) string {
```
mmatczuk, line 34: How about implementing Stringer in frame.Inet

transport/cluster.go
```diff
@@ -0,0 +1,268 @@
...
+	return string(i.IP) + ":" + strconv.Itoa(int(i.Port))
+}
+
+const (
+	nodeAddr   = 0
+	nodeDC     = 1
+	nodeRack   = 2
+	nodeTokens = 3
```
mmatczuk, line 42: What's that?

transport/cluster.go
```diff
@@ -0,0 +1,268 @@
...
+
+	statusDown = 0
+	statusUP   = 1
```
mmatczuk, line 45: Maybe a type?

transport/cluster.go
```diff
@@ -0,0 +1,268 @@
...
+
+	closeRefresher  = 0
+	refreshTopology = 1
```
mmatczuk, line 48: ?

transport/cluster.go
```diff
@@ -0,0 +1,268 @@
...
+)
+
+type PeerMap = map[string]*Node
+
+type Cluster struct {
+	Peers atomic.Value // PeerMap
+
+	Control   ControlConn
+	Refresher chan byte
+
```
mmatczuk, line 58: Should we export the fields?

transport/cluster.go
mmatczuk: Don't cheat, let's have a properly typed internal events

mmatczuk:
```
func (c *Cluster) Refresh() {
 	c.Refresher <- refreshTopology
 }

 func (c *Cluster) StopRefresher() {
 	c.Refresher <- closeRefresher
 }
```

It looks like it should be channel of struct{} and we should be closing it.

session.go
```diff
@@ -0,0 +1,56 @@
+package main // Probably not the best name.
```
mmatczuk, line 1: The right name is scylla!

session.go
```diff
@@ -0,0 +1,56 @@
...
+	cluster            *transport.Cluster
+	defaultConsistency frame.Consistency
+}
+
+type SessionConfig struct {
+	hosts []string // List of hosts given by user at the beginning.
+
+	tcpNoDelay         bool
+	timeout            time.Duration
+	defaultConsistency frame.Consistency
```
mmatczuk, line 21: Here fils should be exported

session.go
```diff
@@ -0,0 +1,56 @@
...
+)
+
+type Session struct {
+	cluster            *transport.Cluster
+	defaultConsistency frame.Consistency
+}
+
+type SessionConfig struct {
+	hosts []string // List of hosts given by user at the beginning.
+
```
mmatczuk, line 18: No grouping pls

session.go
```diff
@@ -0,0 +1,56 @@
...
+	tcpNoDelay         bool
```
mmatczuk, line 19: Let's use ConnConfig

mmatczuk: We can embed it like

```
type SessionConfig struct {
    transport.ConnConfig
```

to make it flat

session.go
```diff
@@ -0,0 +1,56 @@
...
+	return transport.ConnConfig{
+		TCPNoDelay:         s.tcpNoDelay,
+		Timeout:            s.timeout,
+		DefaultConsistency: s.defaultConsistency,
+	}
+}
+
+// Connect creates session and discovers cluster topology via control connection.
+// nolint:unused // This will be used.
+func Connect(cfg SessionConfig) (*Session, error) {
```
mmatczuk, line 35: Why not NewSession?

session.go
```diff
@@ -0,0 +1,56 @@
...
+	if len(cfg.hosts) == 0 {
+		return nil, fmt.Errorf("connecting session requireas at least 1 host address")
```
mmatczuk, line 37: creating

session.go
```diff
@@ -0,0 +1,56 @@
...
+	}
+
+	connCfg := cfg.newConnConfig()
+	cld := &transport.ClusterData{}
+
+	for _, v := range cfg.hosts {
```
mmatczuk, line 43: Nope.

transport/cluster.go
```diff
@@ -0,0 +1,268 @@
...
+	c.SetPeers(m)
+	return nil
+}
+
+type ControlConn struct {
+	Conn       *Conn
+	Cfg        ConnConfig
+	Events     responseHandler
+	KnownHosts []string
+}
```
mmatczuk, line 137: ControlConn can be created from a dialed connection i.e. there should be `NewControlConn(conn Conn)`.

It adds to Conn ability to run some common queries and react to events.

mmatczuk: Fields must not be exported.

transport/cluster.go
```diff
@@ -0,0 +1,268 @@
...
+				Cfg:        cfg,
+				KnownHosts: kh,
+			}, nil
+		}
+	}
+
+	return ControlConn{}, fmt.Errorf("couldn't open control connection to any known host")
+}
+
+func (c ControlConn) RegisterEvents() error {
```
mmatczuk, line 157: Register MUST be moved to Conn, the function should accept event types.

transport/cluster.go
```diff
@@ -0,0 +1,268 @@
...
+	}
+	if _, ok := res.(*Ready); !ok {
+		return fmt.Errorf("invalid response for register request")
+	}
+
+	return nil
+}
+
+// HandleEvents is a loop function which
+func (c *Cluster) HandleEvents() {
```
mmatczuk, line 181: Needs to be refactored to methods for each event type.

transport/node.go
```diff
@@ -0,0 +1,25 @@
...
+
+func InitNodeConnPool(addr string, cfg ConnConfig) *ConnPool {
+	return (*ConnPool)(nil)
+}
+
+type Node struct {
+	addr       string
+	datacenter string
+	rack       string
+	tokens     []byte // It should be []Token, but we don't have parsing for that yet.
```
mmatczuk, line 17: For a DTO fields should be exported...

mmatczuk:
ATM it's quite a mess. 

It looks like you are trying to put too much things in one chunk of work and you got lost in your own balagan.

mmatczuk:
All that session stuff is not needed here.

Let's just add the Cluster + events handling, it seems that the control conn can be just Conn.
We would send queries to control from Cluster.

Please do think about exported vs not exported fields.
Please respect object boundaries!!

transport/cluster.go
Michal-Leszczynski: Wouldn't it be confusing to name it almost the same as the above var peerQuery?

transport/cluster.go
Michal-Leszczynski: These represent specific indexes of columns returned from queries necessary for refresh topology. I will comment it so it becomes more readable.

transport/cluster.go
Michal-Leszczynski: No. Session will interact with them via specific functions.

transport/stream.go
```diff
@@ -9,6 +9,10 @@ import (
 )
 
 const (
+	// There is a problem with StreamID range.
+	// In gocql binary protocol it says that StreamID is a [short] which is unsigned value.
+	// On the other hand all event responses have StreamID equal to -1.
+	// Should -1 be treated as math.MaxInt16?
```
Choraden, line 15: I think it should be MaxUInt16.

go.sum
```diff
@@ -1,3 +1,14 @@
...
 github.com/google/go-cmp v0.5.6 h1:BKbKCqvP6I+rmFHt06ZmyQtvB8xAkWdhFyr0ZUNZcxQ=
 github.com/google/go-cmp v0.5.6/go.mod h1:v8dTdLbMG2kIc/vJvl+f65V22dbkXbowE6jgT/gNBxE=
+github.com/pmezard/go-difflib v1.0.0 h1:4DBwDE0NGyQoBHbLQYPwSUPoCMWR5BEzIk/f1lZbAQM=
+github.com/pmezard/go-difflib v1.0.0/go.mod h1:iKH77koFhYxTK1pcRnkKkqfTogsbg7gZNVY4sRDYZ/4=
+github.com/stretchr/objx v0.1.0/go.mod h1:HFkY916IF+rwdDfMAkV7OtwuqBVzrE8GR6GFx+wExME=
+github.com/stretchr/testify v1.3.0 h1:TivCn/peBQ7UY8ooIcPgZFpTNSz0Q2U6UrFlUfqbe0Q=
+github.com/stretchr/testify v1.3.0/go.mod h1:M5WIy9Dh21IEIfnGCwXGc5bZfKNJtfHm1UVUgZn+9EI=
+go.uber.org/atomic v1.9.0 h1:ECmE8Bn/WFTYwEW/bpKD3M8VtR/zQVbavAoalC1PYyE=
+go.uber.org/atomic v1.9.0/go.mod h1:fEN4uk6kAWBTFdckzkM89CLk9XfWZrxpCo0nPH17wJc=
+golang.org/x/xerrors v0.0.0-20191204190536-9bdfabe68543 h1:E7g+9GITq07hpfrRu66IVDexMakfv52eLZ2CXBWiKr4=
```
Neofine, line 13: Are all of those required?

transport/cluster.go
```diff
@@ -0,0 +1,252 @@
...
+
+func (c *Cluster) OpenControl() error {
+	// TODO: should we random shuffle known hosts after establishing connection?
+	for _, v := range c.knownHosts {
+		if control, err := OpenConn(v, nil, c.cfg); err == nil {
+			c.control = control
+		}
+	}
+
+	return fmt.Errorf("couldn't open control connection to any known host")
```
Neofine, line 78: Should it be the only return of this function?

transport/cluster.go
```diff
@@ -0,0 +1,252 @@
...
+	}
+}
+
+// RequestRefresh notifies cluster that it should update topology.
+// TODO: do we need some mechanism for notifying requester that refresh was successful?
+func (c *Cluster) RequestRefresh() {
+	c.refresher <- struct{}{}
+}
+
+func (c *Cluster) StopEvenHandling() {
```
Neofine, line 207: I guess it should be `StopEventHandling`

transport/cluster.go
```diff
@@ -0,0 +1,252 @@
...
+	close(c.events)
+}
+
+const (
+	refreshChanSize = 32
+	refreshInterval = 60 * time.Second
+)
+
+// Loop handles refresh topology requests.
+func (c *Cluster) Loop() {
```
Neofine, line 217: Should `Loop()` be exported?

transport/cluster.go
```diff
@@ -0,0 +1,252 @@
...
+	c.refresher = make(refreshHandler, refreshChanSize)
+	ticker := time.NewTimer(refreshInterval)
+	for {
+		select {
+		case _, ok := <-c.refresher:
+			if ok {
+				c.tryRefresh()
+				ticker.Stop()
```
Neofine, line 225: Should ticker be stopped right before its reset?

transport/conn.go
```diff
@@ -343,3 +344,19 @@ func (c *Conn) sendRequest(req frame.Request, compress, tracing bool) (frame.Res
 
 	return resp.Response, resp.Err
 }
+
+func (c *Conn) registerEvents(e []frame.EventType) (responseHandler, error) {
+	h := make(responseHandler)
```
Neofine, line 349: I can't grasp the need of doing this, doesn't `sendRequest()` make a responseHandler for you?

transport/conn.go
```diff
@@ -343,3 +344,19 @@ func (c *Conn) sendRequest(req frame.Request, compress, tracing bool) (frame.Res
...
+	c.r.h[eventStreamID] = h
+	req := &Register{EventTypes: e}
+
+	res, err := c.sendRequest(req, false, false)
+	if err != nil {
+		return nil, fmt.Errorf("registering for events: %w", err)
+	}
+	if _, ok := res.(*Ready); !ok {
```
Neofine, line 357: To keep consistency with
https://github.com/mmatczuk/scylla-go-driver/blob/main/transport/conn.go#L354
and
https://github.com/mmatczuk/scylla-go-driver/blob/main/transport/conn.go#L343
can we make this
`if _, ok := res.(*Ready); ok {`
with changed subsequent returns?

transport/node.go
```diff
@@ -0,0 +1,46 @@
...
+	case frame.Up:
+		n.status.Store(statusUP)
+	case frame.Down:
+		n.status.Store(statusDown)
+	default:
+		log.Fatalf("status change not supported: %s", status)
+	}
+}
+
+func (n *Node) GetStatus() bool {
```
Neofine, line 44: If we have a function `GetStatus()` shouldn't we have, even not exported one, storeStatus?

```go
func (n *Node) storeStatus(status bool) {
    n.status.Store(status)
}
```
to keep consistency?

transport/cluster.go
Michal-Leszczynski: Documentation: "For a Timer created with NewTimer, Reset should be invoked only on stopped or expired timers with drained channels".

transport/cluster.go
```diff
@@ -0,0 +1,288 @@
...
+
+type PeerMap = map[string]*Node
+
+type refreshHandler chan struct{}
+
+type Cluster struct {
+	peers         atomic.Value // PeerMap
+	control       *Conn
+	cfg           ConnConfig
+	events        responseHandler // TODO: should events have their own type of chan or is responseHandler fine?
```
Choraden, line 24: We can make alias, this would make it more readable. I would go for it :)

transport/cluster.go
```diff
@@ -0,0 +1,288 @@
...
+	handledEvents []frame.EventType
+	knownHosts    []string // TODO: do we want to update known hosts with node info from cluster?
```
Choraden, line 26: I understand this is the list of host for control connection. Let's keep this code simple as it already is quite a big pr.
We can definitely add this kind of feature later.

transport/cluster.go
```diff
@@ -0,0 +1,288 @@
...
+		return nil, fmt.Errorf("creating cluster: %w", err)
+	}
+
+	go c.handleEvents()
+	go c.loop()
+	return c, nil
+}
+
+func (c *Cluster) OpenControl() error {
+	// TODO: should we random shuffle known hosts after establishing connection?
```
Choraden, line 70: I don't find it helpful.

transport/conn.go
```diff
@@ -437,3 +438,19 @@ func (c *Conn) Close() {
 func (c *Conn) String() string {
 	return fmt.Sprintf("[addr=%s shard=%d]", c.conn.RemoteAddr(), c.shard)
 }
+
+func (c *Conn) registerEvents(e []frame.EventType) (responseHandler, error) {
+	h := make(responseHandler, 2)
+	c.r.h[eventStreamID] = h
```
Choraden, line 444: Although there is only one writer to this connection, I would do it under mutex to keep it clean. What do you think?

transport/cluster.go
```diff
@@ -0,0 +1,288 @@
...
+const defaultPort = 9042
+
+func appendDefaultPort(addr string) string {
+	return addr + ":" + strconv.Itoa(defaultPort)
+}
+
+const eventChanSize = 32
+
+func (c *Cluster) handleEvents() {
+	c.events = make(responseHandler, eventChanSize)
```
Choraden, line 181: I don't understand we would me make channel here and also in registerEvents?

transport/node.go
```diff
@@ -0,0 +1,29 @@
...
+	statusUP   = true
+)
+
+type Node struct {
+	addr       string
+	datacenter string
+	rack       string
+	tokens     []byte // It should be []Token, but we don't have parsing for that yet.
+	pool       *ConnPool
+	status     *nodeStatus
```
Choraden, line 20: Why pointer here?

transport/cluster.go
```diff
@@ -0,0 +1,288 @@
...
+			ticker.Reset(refreshInterval)
+		}
+	}
+}
+
+// tryRefresh refreshes cluster topology.
+// In case of error tries to reopen control connection and tries again.
+func (c *Cluster) tryRefresh() {
+	if err := c.refreshTopology(); err != nil {
+		if err = c.OpenControl(); err != nil {
```
Choraden, line 262: Shouldn't we firstly close currently opened control connection?

transport/cluster.go
Michal-Leszczynski: I found something similar in rust so I wanted to have second opinion on this.

transport/conn.go
Michal-Leszczynski: If we want to do it with mutex (which is probably not necessary, but won't effect efficiency), we should wrap it in some method. I will try to implement this approach.

transport/cluster.go
Michal-Leszczynski: It's a bug, thanks for catching it.

transport/node.go
Michal-Leszczynski: "go.uber.org/atomic" package implements methods on pointers, like:
    func NewBool(val bool) *atomic.Bool
    func (x *Bool) Load() bool

So we need pointers. The question is, is it better to have:
    type nodeStatus = atomic.Bool
    type Node struct {
        ...
        status *nodeStatus
    }
or
    type nodeStatus = *atomic.Bool
    type Node struct {
        ...
        status nodeStatus
    }

transport/conn.go
Choraden: thanks!

frame/response/event.go
```diff
@@ -71,3 +72,18 @@ func ParseSchemaChange(b *frame.Buffer) *SchemaChange {
...
+	s := b.ReadString()
+	switch s {
+	case "TOPOLOGY_CHANGE":
+		return ParseTopologyChange(b)
+	case "STATUS_CHANGE":
+		return ParseStatusChange(b)
+	case "SCHEMA_CHANGE":
+		return ParseSchemaChange(b)
+	default:
+		log.Fatalf("event type not supported: %s", s)
```
mmatczuk, line 86: This is not how we handle errors here, this would exit the APP for no good reason.

mmatczuk: I find lot's of this.

```
	if Debug {
		log.Printf("unknown consistency: %v", v)
	}
```

transport/cluster.go
```diff
@@ -0,0 +1,295 @@
...
+	. "scylla-go-driver/frame/response"
+
+	"go.uber.org/atomic"
+)
+
+type PeerMap = map[string]*Node
+
+type eventHandler = responseHandler
+
+type refreshHandler chan struct{}
```
mmatczuk, line 20: FYI it's possible to use grouping on types too, not a must here.

transport/stream.go
```diff
@@ -9,7 +9,8 @@ import (
 )
 
 const (
-	maxStreamID = math.MaxInt16
+	maxStreamID   = math.MaxInt16
+	eventStreamID = math.MaxUint16 // TODO: is this correct?
```
mmatczuk, line 13: It's seems utterly wrong

```
  A frame has a stream id (a [short] value). When sending request messages, this
  stream id must be set by the client to a non-negative value (negative stream id
  are reserved for streams initiated by the server; currently all EVENT messages
  (section 4.2.6) have a streamId of -1). If a client sends a request message
  with the stream id X, it is guaranteed that the stream id of the response to
  that message will be X.
```

- stream ID is 2 bytes singed int i.e. Int16
- EVENTS use negative IDs

mmatczuk: Apparently `ffff` is binary coding of both `-1` as int16 and `math.MaxUint16`...

Can we use int16(-1)?

transport/node.go
```diff
@@ -0,0 +1,29 @@
...
+const (
+	statusDown = false
+	statusUP   = true
+)
+
+type Node struct {
+	Addr       string
+	Datacenter string
+	Rack       string
+	Tokens     []byte // It should be []Token, but we don't have parsing for that yet.
```
mmatczuk, line 18: Let's add TODO comments for all things that needs to be added immediately after

transport/node.go
```diff
@@ -0,0 +1,29 @@
+package transport
+
+import (
+	"go.uber.org/atomic"
+)
+
+type nodeStatus = *atomic.Bool
```
mmatczuk, line 7: The pointer here is not needed, unless all APIs use Node by copy.
From what I see we have *Node, in that case compiler takes pointer from the struct on heap, no need for a dedicated pointer here.

transport/node.go
```diff
@@ -0,0 +1,29 @@
...
+
+const (
+	statusDown = false
+	statusUP   = true
+)
+
+type Node struct {
```
mmatczuk, line 14: @Michal-Leszczynski 

- The idiomatic go naming scheme for field `x` is `X` and `SetX` no need for Get
- IMO all fields here should NOT be exported but they should have getters, this is API for main package and it prevents from untrusted code to change it...
- Also SetStatus should not be exported, I'd consider removing it

transport/conn.go
```diff
@@ -437,3 +447,18 @@ func (c *Conn) Close() {
 func (c *Conn) String() string {
 	return fmt.Sprintf("[addr=%s shard=%d]", c.conn.RemoteAddr(), c.shard)
 }
+
+func (c *Conn) registerEvents(e []frame.EventType) (eventHandler, error) {
+	h := c.r.setEventHandler()
```
mmatczuk, line 452: I fail to understand why the functions are so distant in the code?
They should go one after the other setEventHandler being after registerEvents, it's a helper function.

transport/conn.go
```diff
@@ -437,3 +447,18 @@ func (c *Conn) Close() {
 func (c *Conn) String() string {
 	return fmt.Sprintf("[addr=%s shard=%d]", c.conn.RemoteAddr(), c.shard)
 }
+
+func (c *Conn) registerEvents(e []frame.EventType) (eventHandler, error) {
```
mmatczuk, line 451: You could make the API nicer like `registerEvents(e... frame.EventType)` it would auto make the slice.

transport/conn.go
```diff
@@ -437,3 +447,18 @@ func (c *Conn) Close() {
...
+	h := c.r.setEventHandler()
+	req := &Register{EventTypes: e}
```
mmatczuk, line 453: You can inline it.

transport/conn.go
```diff
@@ -437,3 +447,18 @@ func (c *Conn) Close() {
...
+
+	res, err := c.sendRequest(req, false, false)
+	if err != nil {
+		return nil, fmt.Errorf("registering for events: %w", err)
+	}
+	if _, ok := res.(*Ready); ok {
```
mmatczuk, line 459: Else there is `responseAsError` function

transport/cluster.go
```diff
@@ -0,0 +1,295 @@
...
+
+	"scylla-go-driver/frame"
+	. "scylla-go-driver/frame/response"
+
+	"go.uber.org/atomic"
+)
+
+type PeerMap = map[string]*Node
+
+type eventHandler = responseHandler
```
mmatczuk, line 18: This can be inlined / removed.

transport/cluster.go
```diff
@@ -0,0 +1,295 @@
...
+}
+
+// type dataCenter struct {
+// 	nodes   []Node
+// 	rackCnt uint // Type?
+// }
+
+// NewCluster also creates control connection and starts handling events and refreshing topology.
+func NewCluster(cfg ConnConfig, e []frame.EventType, hosts ...string) (*Cluster, error) {
+	if len(hosts) < 1 {
```
mmatczuk, line 43: Typically == 0

transport/cluster.go
```diff
@@ -0,0 +1,295 @@
...
+		return nil, fmt.Errorf("at least one host is required to create cluster")
+	}
+
+	p := atomic.Value{}
+	p.Store(PeerMap{})
+	k := make([]string, len(hosts))
+	copy(k, hosts)
```
mmatczuk, line 50: 👍

transport/cluster.go
```diff
@@ -0,0 +1,295 @@
...
+
+	c := &Cluster{
+		peers:         p,
+		cfg:           cfg,
+		handledEvents: e,
+		knownHosts:    k,
+	}
+
+	if err := c.NewControl(); err != nil {
+		return nil, fmt.Errorf("creating cluster: %w", err)
```
mmatczuk, line 60: I'd refrain from `ing` active voice in error messages.
Plain dull `create cluster: open X: bla` is shorter and reads better.
@Michal-Leszczynski this is a general comment to all code here.

mmatczuk: technically it should be `create control connection` or sth.

transport/cluster.go
```diff
@@ -0,0 +1,295 @@
...
+	}
+	if err := c.refreshTopology(); err != nil {
+		return nil, fmt.Errorf("creating cluster: %w", err)
```
mmatczuk, line 63: this one is `refresh topology`

mmatczuk: See the pattern?

transport/cluster.go
```diff
@@ -0,0 +1,295 @@
...
+	}
+
+	go c.loop()
+	return c, nil
+}
+
+func (c *Cluster) NewControl() error {
```
mmatczuk, line 70: Logging, log error messages with  host address!

transport/cluster.go
```diff
@@ -0,0 +1,295 @@
...
+	log.Printf("opening control connection")
+	for _, v := range c.knownHosts {
```
mmatczuk, line 72: How about addr ?

transport/cluster.go
```diff
@@ -0,0 +1,295 @@
...
+		control, err := OpenConn(v, nil, c.cfg)
```
mmatczuk, line 73: How about conn?

transport/cluster.go
```diff
@@ -0,0 +1,295 @@
...
+
+				go c.handleEvents()
+				return nil
+			}
+		}
+		if control != nil {
+			control.Close()
+		}
+	}
+	return fmt.Errorf("couldn't open control connection to any known host")
```
mmatczuk, line 87: Let's add the list of know hosts here.

transport/cluster.go
```diff
@@ -0,0 +1,295 @@
...
+}
+
+// refreshTopology creates new PeerMap filled with the result of both localQuery and peerQuery.
+// The old map is replaced with the new one atomically to prevent dirty reads.
+func (c *Cluster) refreshTopology() error {
+	log.Printf("refreshing topology")
+	rows, err := c.getAllNodesInfo()
+	if err != nil {
+		return fmt.Errorf("refreh topology: %w", err)
```
mmatczuk, line 96: Again errors messaged describe what you tried not where you are.

transport/cluster.go
```diff
@@ -0,0 +1,295 @@
...
+	}
+	// If node is present in both maps we can update already created node
+	// instead of creating new one from the scratch.
+	old := c.GetPeers()
+	m := PeerMap{}
+	for _, v := range rows {
+		addr := net.IP(v[nodeAddr]).String()
```
mmatczuk, line 103: You may add a TODO as this should probably change when we have nicer row handling.

transport/cluster.go
```diff
@@ -0,0 +1,295 @@
...
+			node.Tokens = v[nodeTokens]
+			m[addr] = node
+		} else {
+			n := &Node{
+				Addr:       addr,
+				Datacenter: string(v[nodeDC]),
+				Rack:       string(v[nodeRack]),
+				Tokens:     v[nodeTokens],
+			}
+			if pool, err := NewConnPool(appendDefaultPort(addr), c.cfg); err != nil {
```
mmatczuk, line 115: Please do not append it here.
It should be appended in NewConnPool!

mmatczuk: There is some discussion related to that [in this PR comment](https://github.com/mmatczuk/scylla-go-driver/pull/148#pullrequestreview-906906584).

transport/cluster.go
```diff
@@ -0,0 +1,295 @@
...
+				n.Status = atomic.NewBool(statusDown)
+			} else {
+				n.Status = atomic.NewBool(statusUP)
```
mmatczuk, line 118: This forces additional allocations, pointer not needed.

transport/cluster.go
```diff
@@ -0,0 +1,295 @@
...
+			m[addr] = n
+		}
+	}
+
+	for k, v := range old {
+		if _, ok := m[k]; !ok {
+			v.Pool.Close()
+		}
+	}
+	c.SetPeers(m)
```
mmatczuk, line 130: SetPeers should not be exported.

transport/cluster.go
```diff
@@ -0,0 +1,295 @@
...
+
+	localRes, err := c.control.Query(localQuery, nil)
+	if err != nil {
+		return nil, fmt.Errorf("discovering local topology: %w", err)
+	}
+
+	return append(peerRes.Rows, localRes.Rows[0]), nil
+}
+
+func (c *Cluster) GetPeers() PeerMap {
```
mmatczuk, line 168: Just Peers

transport/cluster.go
```diff
@@ -0,0 +1,295 @@
...
+}
+
+const (
+	refreshChanSize = 32
+	refreshInterval = 60 * time.Second
+)
+
+// loop handles refresh topology requests.
+func (c *Cluster) loop() {
+	c.refresher = make(refreshHandler, refreshChanSize)
```
mmatczuk, line 244: This should be done in constructor.

transport/cluster.go
```diff
@@ -0,0 +1,295 @@
...
+		case *SchemaChange:
+			// TODO: add schema change.
+		default:
+			log.Printf("unsupported event type: %v", res.Response)
+		}
+	}
+}
+
+func (c *Cluster) handleTopologyChange(v *TopologyChange) {
+	log.Printf("handling node: %s topology change to: %s", v.Address.String(), v.Change)
```
mmatczuk, line 204: Log the whole thing `%+#v`

transport/cluster.go
```diff
@@ -0,0 +1,295 @@
...
+	c.RequestRefresh()
+}
+
+func (c *Cluster) handleStatusChange(v *StatusChange) {
+	log.Printf("handling node: %s status change to: %s", v.Address.String(), v.Status)
```
mmatczuk, line 209: Log the whole thing %+#v

transport/cluster.go
```diff
@@ -0,0 +1,295 @@
...
+	m := c.GetPeers()
+	addr := v.Address.String()
+	if n, ok := m[addr]; ok {
+		switch v.Status {
+		case frame.Up:
+			n.SetStatus(statusUP)
+		case frame.Down:
+			n.SetStatus(statusDown)
+		default:
+			log.Fatalf("status change not supported: %s", v.Status)
```
mmatczuk, line 219: Clearly this should be error logged but fatal would kill the whole APP.

transport/cluster.go
```diff
@@ -0,0 +1,295 @@
...
+		}
+	} else {
+		log.Printf("node which status is being set to %s was not present in known topology: %s", v.Status, addr)
```
mmatczuk, line 222: This log line is too long and does not print the topology but add

transport/cluster.go
```diff
@@ -0,0 +1,295 @@
...
+		c.RequestRefresh()
+	}
+}
+
+// RequestRefresh notifies cluster that it should update topology.
+// TODO: do we need some mechanism for notifying requester that refresh was successful?
```
mmatczuk, line 228: No.

transport/cluster.go
```diff
@@ -0,0 +1,295 @@
...
+func (c *Cluster) RequestRefresh() {
+	c.refresher <- struct{}{}
+}
+
+func (c *Cluster) StopEventHandling() {
```
mmatczuk, line 233: Should't this be called Close?

mmatczuk: Is there StartEventHandling

mmatczuk: @Michal-Leszczynski there is a fundamental BUG(s) here!!

- handleEvents is not prepared for this to close
- you are closing something you do not own
- conn will panic when you close it

mmatczuk (changes requested):
Left some comments inline.

One more general comment use `if err := ` instead of `if err = `.

There is major flaw in the closing workflow.

Namely if `c.control.Close()` there is a go routine leak and calling StopEventHandling does not really stop anything can can cause panic when conn tries to handle events.

```
	if err := c.refreshTopology(); err != nil {
		log.Printf("failed to refresh topology: %s", err.Error())
		c.control.Close()
```

# My suggestion

- cluster should have Close() function that would effectively close everything - all the Stop this stop that should not be exported or even removed
- cluster::events should be created in NewCluster
- cluster::events should be passed to Conn::registerEvents as input parameter - it's fully operated by cluster not conn - better idea later
- connReader need special handling of event stream because putting to cluster::events can block
- it would be better to pass a function to Conn::registerEvents that would be called on EVENT instead of registering a handler like in normal request sending
- with that there is a real full control of cluster::events in Cluster itself
- when you put to cluster::events it should be done via [select with default](https://golangbyexample.com/select-default-case-go/) namely if event processing is too slow it allows to DROP the event log it and DO NOT BLOCK connReader

Otherwise good piece of code.

Looking forward to next version.

transport/cluster.go
Michal-Leszczynski: That's true and I actually spotted this bug because of TestClusterIntegration!
I already fixed it, but forgot to delete this useless function.

frame/response/event.go
Michal-Leszczynski: For now Debug is constant set in frame/debug.go. Should we add another debug constant or make one for all packages?

transport/cluster.go
mmatczuk: See the final statement https://github.com/mmatczuk/scylla-go-driver/pull/130#pullrequestreview-907045101

transport/stream.go
Michal-Leszczynski:
```
To describe the layout of the frame body for the messages in Section 4, we define the following:
    [int]            A 4 bytes integer
    [long]         A 8 bytes integer
    [short]        A 2 bytes unsigned integer
    ...
```
Since here short is described as unsigned, we were hesitating with the type. In frame we have:
```
// StreamID is a type alias for Short.
type StreamID = Short
```
and in transport we have:
```
type connReader struct {
	...
	h      map[frame.StreamID]responseHandler
        ...
}
```
So if we use **int16(-1)** we would still require casting it to **uint16** in other parts of code.

Can we stay with **math.MaxUint16** and mark it with appropriate comment?
Or should we change the rest of the code so that StreamID is of type **int16**, because perhaps that's a little bug in **gocql binary protocol**?

BTW:
`maxStreamID = math.MaxInt16`
So it is probably doable without greater changes.

mmatczuk: This is very interesting!
[Here]https://github.com/apache/cassandra/blob/2b2c6decfafc6235ad537e72073fab2fd4467e2f/doc/native_protocol_v4.spec#L160) they clearly state that those 2 bytes
- can can hold up to 0x800 (max int 16 singed)
- -1 is EVENTS stream

I think that we can change 

```
// StreamID is a type alias for Short.
type StreamID = int16

// Header spec https://github.com/apache/cassandra/blob/trunk/doc/native_protocol_v4.spec#L101.
type Header struct {
...
	StreamID StreamID
}
```

transport/node.go
Michal-Leszczynski: > For a DTO fields should be exported...
I got confused by your previous comment, but I agree with this approach.

transport/conn.go
Michal-Leszczynski: I wanted to have **setEventHandler** next to **setHandler**, but moving it makes more sense.

Michal-Leszczynski:
Im not sure whether it's worth to worry about blocking when putting response into cluster::events channel. I would assume that events are rarely received and the chance of blocking the channel is really small (we can set channel size to 1024 and then I would imagine that's pretty much impossible to block it in regular conditions).

But passing cluster::events down to connection and resigning from registering handler seems like a good idea and I will proceed with this approach (and with special event handling in connReader as well)

Michal-Leszczynski:
Currently we have to separate goroutines in cluster: **loop** and **handleEvents**, but for now handleEvents only delegates work to **loop** or does some minor changes in cluster so it does not seem like it's that important?
In the future we will need to request some metadata from the cluster (also connected to schema change event) and we would like to refresh this data in terms of error or encountering some unknown tables or keyspaces, but this task does not really fall under **handleEvent** jurisdiction.

My idea is to have another loop responsible for handling refresh metadata requests, just like the **loop** responsible for handling topology refresh requests. In this scenario handling events is done just by passing some function (which will make some minor changes in cluster or delegate work directly to one of those two loops) to register events.

Another general problem arises from the fact, that if we have two goroutines and each of them wants to reopen control connection in case of error, we have to synchronize it. I see three possible solutions:
- some mutex guarding connection
- making two connections, one for each loop
- making one loop handle reopen control connection request (then one loop will be delegating this work to another)
the last one may seem a little out of place, but is also the least problematic one. 

I will try to read more rust to have a better perspective, but I would like to hear your opinion on this approach.

mmatczuk:
@Michal-Leszczynski 

Regarding the passing cluster::events 
- note that the issue of blocking exists and it needs to be sent with select
- please do not make the buffer too deep

Regarding go routines, it is best to have just one (the loop) function and use select to read from many channels.
There is not need for any synchronization then, it's synchronized by message passing.

mmatczuk:
I had a look at the commit with the new approach.

- You can use `c.handleEvent` as input, there is no need to necessarily return the new function
- I like the idea with type of action because there is a single channel then - the code is easier
- Yet still Conn and Cluster should communicate by message passing

mmatczuk:
Overall I think that many channels and single loop is more idiomatic and easier to maintain.

Michal-Leszczynski:
With many channels we can set their buff length to 1, so that multiples of the same request (like refresh topology) will be handled by just one refresh topology action.

mmatczuk:
That is correct.

Michal-Leszczynski:
Corrected, squashed and rebased onto main version is on **ml/delete_evet_loop** PR. Can we merge it from there?

mmatczuk:
Yes.

PR #131, opened 2022-02-25
session: basic usability

Choraden:


session.go
```diff
@@ -0,0 +1,63 @@
...
+)
+
+type Session struct {
+	cfg     SessionConfig
+	cluster transport.Cluster
+}
+
+type Result = transport.QueryResult
+
+func NewSession(config SessionConfig, hosts ...string) (*Session, error) {
```
mmatczuk, line 18: Why hosts are not in SessionConfig?

mmatczuk:
@Choraden

Choraden:
@Kulezi please tell me what you think of this, especially aliasing Statement and Result. 
I think this solution has some pros, since Statement is really close to Query abstraction.
You could add Iter to QueryResult as well.

session.go
```diff
@@ -0,0 +1,113 @@
...
+
+import (
+	"fmt"
+	"net"
+	"time"
+
+	"scylla-go-driver/transport"
+)
+
+// TODO: Should we support connecting to different than default port?
```
mmatczuk, line 11: Yes but not now open an issue about it.
The host can be just the host or the full address.
p3.

transport/node.go
```diff
@@ -29,3 +29,7 @@ func (n *Node) Status() bool {
 func (n *Node) setStatus(v bool) {
 	n.status.Store(v)
 }
+
+func (n *Node) RandomConnection() *Conn {
+	return n.pool.LeastBusyConn()
+}
```
mmatczuk, line 35: Remove it.

query.go
```diff
@@ -0,0 +1,28 @@
...
+		N: 8,
+		Bytes: []byte{
+			byte(v >> 56),
+			byte(v >> 48),
+			byte(v >> 40),
+			byte(v >> 32),
+			byte(v >> 24),
+			byte(v >> 16),
+			byte(v >> 8),
+			byte(v),
```
mmatczuk, line 23: Don't we have a helper for that? Note the [ByteOrder](https://pkg.go.dev/encoding/binary).

session.go
```diff
@@ -0,0 +1,113 @@
...
+// TODO: Add retry policy.
+// TODO: Add Query Paging.
+// TODO: make better randomConnection.
+
+type SessionConfig struct {
+	hosts []string
```
mmatczuk, line 17: All config options shall be Exported.

session.go
```diff
@@ -0,0 +1,113 @@
...
+	transport.ConnConfig
+}
+
+func MakeSessionConfig(tcpNoDelay bool, defaultConsistency uint16, hosts ...string) SessionConfig {
```
mmatczuk, line 21: Config object are constructed manually as structs, please remove.
One thing to note here is making the default (zero) values useful.

session.go
```diff
@@ -0,0 +1,113 @@
...
+func discoverHosts(hosts []string) []net.IP { // nolint:deadcode,unused // This might be used in the future.
+	var res []net.IP
+	for _, h := range hosts {
+		if ip, err := net.LookupIP(h); err != nil {
+			res = append(res, ip[0])
+		}
+	}
+	return res
+}
+
```
mmatczuk, line 45: As discussed we want to use the host names (DNS) down to the final dial.
Please remove.

session.go
mmatczuk: Let's add Events and Keyspace here too.

session.go
```diff
@@ -0,0 +1,113 @@
...
+		ConnConfig: transport.ConnConfig{
+			TCPNoDelay:         tcpNoDelay,
+			Timeout:            5 * time.Second, // Should user have control of timeout value?
+			DefaultConsistency: defaultConsistency,
+		},
+	}
+}
+
+// I don't know if we need this.
+func discoverHosts(hosts []string) []net.IP { // nolint:deadcode,unused // This might be used in the future.
```
mmatczuk, line 36: As discussed we want to use the host names (DNS) down to the final dial.
Please remove.

session.go
```diff
@@ -0,0 +1,113 @@
...
+	return res
+}
+
+type Session struct {
+	cfg     SessionConfig
+	cluster *transport.Cluster
+}
+
+func NewSession(cfg SessionConfig) (*Session, error) {
+	cluster, err := transport.NewCluster(cfg.ConnConfig, nil, cfg.hosts...)
```
mmatczuk, line 52: Events

session.go
```diff
@@ -0,0 +1,113 @@
...
+
+	s := &Session{
+		cfg:     cfg,
+		cluster: cluster,
+	}
+
+	return s, nil
+}
+
+func (s *Session) NewQuery(content string) Query {
```
mmatczuk, line 65: How about `Query(cql string) *Query`?

transport/node.go
```diff
@@ -29,3 +29,7 @@ func (n *Node) Status() bool {
 func (n *Node) setStatus(v bool) {
 	n.status.Store(v)
 }
+
+func (n *Node) RandomConnection() *Conn {
```
mmatczuk, line 33: Let's preserve the naming please...

transport/query.go
```diff
@@ -29,6 +30,19 @@ func newQueryForStatement(s Statement, pagingState frame.Bytes) frame.Request {
 	}
 }
 
+func newQueryForExecute(s Statement, pagingState frame.Bytes) frame.Request {
```
mmatczuk, line 33: A better name?

transport/query.go
```diff
@@ -8,6 +8,7 @@ import (
 
 type Statement struct {
 	Content           string
+	ID                frame.Bytes
```
mmatczuk, line 11: Maybe ID should go first?

transport/conn.go
```diff
@@ -415,6 +415,32 @@ func (c *Conn) Query(s Statement, pagingState frame.Bytes) (QueryResult, error)
 	return makeQueryResult(res)
 }
 
+func (c *Conn) Prepare(s Statement) (Statement, error) {
+	req := &Prepare{Query: s.Content}
```
mmatczuk, line 419: inline?

transport/conn.go
```diff
@@ -415,6 +415,32 @@ func (c *Conn) Query(s Statement, pagingState frame.Bytes) (QueryResult, error)
...
+		return Statement{}, err
+	}
+
+	if v, ok := res.(*PreparedResult); ok {
+		s.ID = v.ID
+		s.Values = make([]frame.Value, len(v.Metadata.Columns))
+		return s, nil
+	}
+
+	return Statement{}, fmt.Errorf("prepare failed")
```
mmatczuk, line 431: Maybe use responseAsError(res)

transport/conn.go
```diff
@@ -415,6 +415,32 @@ func (c *Conn) Query(s Statement, pagingState frame.Bytes) (QueryResult, error)
...
+}
+
+func (c *Conn) Execute(s Statement, pagingState frame.Bytes) (QueryResult, error) {
+	req := newQueryForExecute(s, pagingState)
```
mmatczuk, line 435: inline?

session.go
```diff
@@ -0,0 +1,113 @@
...
+	return Query{stmt: transport.Statement{Content: content, Consistency: s.cfg.DefaultConsistency}}
+}
+
+// TODO: This need to be replaced with proper random, but first let's wait for the node selection mechanism and see how it works.
+func (s *Session) randomConnection() *transport.Conn {
```
mmatczuk, line 70: I has nothing to do with randomness, also the RandomConnection implementation would never return nil.
The connection may be broken and everything but not nil.

mmatczuk:
Looks legit, keep it up.

mmatczuk:
Add support for Async API like in https://github.com/mmatczuk/scylla-go-driver/pull/173

mmatczuk:
In the context of the Async APU I think that the `QueryAsync` should be moved to Session,
and Conn should have the normal (sync) API only.

Please migrate that.

session.go
Choraden: By the way, it seems oddly to me to let somebody see what's inside the driver and what is the structure of internal packages. What is the policy of visibility in such config structs then? Should user see our internal types i.e. frame.EventType or transport.something?

session.go
Choraden: 1. There is Query method which runs everything. What should I name it then? 
```go
func (s *Session) Query(req Query) (Result, error) {
	conn := s.randomConnection()
	if conn == nil {
		return Result{}, errNoConnection
	}

	res, err := conn.Query(req.stmt, nil)
	return Result(res), err
}
```
2. I just want to be sure if the pointer is going to point to query struct on stack, right?

query.go
Choraden: As I look at helper functions from binary package I can see they require slice to have valid length. 
```go
func (bigEndian) PutUint64(b []byte, v uint64) {
	_ = b[7] // early bounds check to guarantee safety of writes below
	b[0] = byte(v >> 56)
	b[1] = byte(v >> 48)
	b[2] = byte(v >> 40)
	b[3] = byte(v >> 32)
	b[4] = byte(v >> 24)
	b[5] = byte(v >> 16)
	b[6] = byte(v >> 8)
	b[7] = byte(v)
}

// PutUvarint encodes a uint64 into buf and returns the number of bytes written.
// If the buffer is too small, PutUvarint will panic.
func PutUvarint(buf []byte, x uint64) int {
	i := 0
	for x >= 0x80 {
		buf[i] = byte(x) | 0x80
		x >>= 7
		i++
	}
	buf[i] = byte(x)
	return i + 1
}
```

I don't like this api, because first of all i want to omit allocation and the second of all I'm not a fan of the flow in which the program creates a slice and then passes it to a function.  It would be nicer if the function could return such slice. 

By creating slice in place (i.e. []byte{byte(x), byte(x >> 8), ...}) I understand we skip allocation, thus I decided to do it this way. 

Please tell me if I don't get it right or I'm missing some function from binary package.

session.go
Choraden: @mmatczuk ping

mmatczuk:
`session: sessionConfig is now passed by pointer Hubert Grochowski Yesterday 21:30` - this is a mistake allows caller to modify it

mmatczuk:
Good work with splitting stuff to commits.

Choraden:
> `session: sessionConfig is now passed by pointer Hubert Grochowski Yesterday 21:30` - this is a mistake allows caller to modify it

Linter suggested me this brilliant idea due to this struct heaviness, but you're right. I'll revert it.

session.go
```diff
@@ -0,0 +1,191 @@
...
+
+	s := &Session{
+		cfg:     cfg,
+		cluster: cluster,
+	}
+
+	return s, nil
+}
+
+func (s *Session) leastBusyConn() *transport.Conn {
```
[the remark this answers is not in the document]
Choraden: Don't worry about it, it's a prop. I'm really looking forward to integrating host selection policy :)

session.go
```diff
@@ -0,0 +1,191 @@
...
+	Events             []EventType
+	TCPNoDelay         bool
+	ConnectionTimeout  time.Duration
+	DefaultConsistency Consistency
+}
+
+func (s *SessionConfig) copy() *SessionConfig {
+	var cfgCopy SessionConfig
+
+	cfgCopy.Hosts = make([]string, len(s.Hosts))
```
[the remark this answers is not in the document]
Choraden: Map seems to be reasonable idea.

mmatczuk:
I rebased, it seems to me that it's impossible to access conn pool @Michal-Leszczynski 

```
// FIXME: to be replaced by host selection policy.
func (s *Session) leastBusyConn() *transport.Conn {
	for _, node := range s.cluster.Peers() {
		conn := node.LeastBusyConn()
		if conn != nil {
			return conn
		}
	}

	return nil
}
```

PR #141, opened 2022-03-05
golangci: turned off default exclusions

Choraden:
Turns out there was a hidden list of issues excluded by default. I turned it off, but errors reported by linter still have to be fixed.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -62,6 +62,11 @@ linters:
     - wsl
 
 issues:
+  exclude-use-default: false
   exclude:
+    - exported
     - indent-error-flow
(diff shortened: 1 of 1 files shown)


mmatczuk:
Why the list of excludes is so long?

Choraden:
I don't find them helpful at this stage.

- ST1000: at least one file in a package should have a package comment
- stdmethods: public methods should return error (e.g. stdmethods: method WriteByte(v byte) should have signature WriteByte(byte) error)
- exported: every exported method should have a comment

mmatczuk:
I'm merging this.

PR #142, opened 2022-03-05
Pool: Fixed commented out tests

### LINKED ISSUE #135: pool: fix commented integration tests

Neofine:
Also added constant to increase readability

Fixes #135

### DIFF
--- a/transport/pool_integration_test.go
+++ b/transport/pool_integration_test.go
@@ -7,6 +7,8 @@ import (
 	"time"
 )
 
+const refillerBackoff = 250 * time.Millisecond
+
 func TestNodeConnPoolIntegration(t *testing.T) {
 	p, err := NewConnPool(TestHost+":9042", ConnConfig{})
 	if err != nil {
@@ -14,7 +16,7 @@ func TestNodeConnPoolIntegration(t *testing.T) {
 	}
 
 	t.Log("Wait for refiller to fill connections to shards")
-	time.Sleep(250 * time.Millisecond)
+	time.Sleep(refillerBackoff)
 
 	t.Log("Check if connections to shards are established")
 	for i, c := range p.AllConns() {
@@ -29,7 +31,7 @@ func TestNodeConnPoolIntegration(t *testing.T) {
 	}
 
 	t.Log("Wait for refiller to fill connections to shards")
-	time.Sleep(250 * time.Millisecond)
+	time.Sleep(refillerBackoff)
 
 	t.Log("Check if connections have been refilled")
 	for i, c := range p.AllConns() {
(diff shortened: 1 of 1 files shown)


transport/pool_integration_test.go
```diff
@@ -40,7 +42,36 @@ func TestNodeConnPoolIntegration(t *testing.T) {
...
+
+	t.Log("Check if connections are closed")
+	for _, c := range p.AllConns() {
+		if c != nil {
+			t.Fatalf("conn %s", c)
+		}
+	}
+}
+
+func TestCloseConnPoolIntegration(t *testing.T) {
```
mmatczuk, line 55: What is the purpose of this test i.e. what does it test?

transport/pool_integration_test.go
```diff
@@ -50,77 +81,52 @@ func TestNodeConnPoolIntegration(t *testing.T) {
...
+	}
+
+	murmurToken = MurmurToken([]byte("0")) // Very big number approx. 3 * 10^18.
+
+	if conn := p.Conn(murmurToken); conn == nil {
+		t.Fatal("invalid return of Conn")
+	}
+
+	go func() {
+		seenNil := false
```
mmatczuk, line 111: Seen nil is useless if you can iterate over all connections.

transport/pool_integration_test.go
```diff
@@ -50,77 +81,52 @@ func TestNodeConnPoolIntegration(t *testing.T) {
...
-//			}
-//		}
-//	}()
-//
-//	cp.Close()
-//
-//	// Waiting for go func().
-//	time.Sleep(1 * time.Second)
-//}
+func TestConnPoolConnIntegration(t *testing.T) {
```
mmatczuk, line 84: Another strange test, what it does?

It would be better to add some token based checks to the 1st test case.

transport/pool_integration_test.go
Neofine: This test assures that no unwanted behavior happens if in between shard close requests there happens pool refiller close request.
This was a bug a while ago which this test helped to solve so I didn't remove it just in case, should it be kept?

transport/pool_integration_test.go
Neofine: I have tried to test here few things:

1. As you have mentioned token shard based checks 
2. The go func() for 100 milliseconds tries to get RandConn while p.Close() is called in the test function. With this test I wanted to make sure that even though successive connections are being closed RandConn is consistent with its returns i.e. there is no case of RandConn returning working connection, next RandConn will return nil and after that RandConn would return another working connection. That would hint that something is wrong with either closing connections or pool refiller still refilling. Maybe it's overthinked but that was my intention.

mmatczuk:
Those tests do not add any value and are removed.

mmatczuk:
I merged the variable backoff change.

PR #143, opened 2022-03-06
Pool: Replaced RandomConn and NextConn with LeastBusyConn

### LINKED ISSUE #137: transport: pool, replace random and next with least busy connection
Relates to https://github.com/scylladb/gocql/pull/86

Neofine:
Inspired by https://github.com/scylladb/gocql/pull/86

Fixes #137


transport/pool.go
```diff
@@ -34,30 +33,34 @@ func NewConnPool(addr string, connConfig ConnConfig) (*ConnPool, error) {
...
+		leastBusyConn *Conn
+		minBusy       = math.MaxUint32
+	)
 
-func (p *ConnPool) NextConn() *Conn {
 	idx := p.idx.Add(1)
 	for i := 0; i < p.nrShards; i++ {
 		if conn := p.loadConn(int((idx + uint64(i)) % uint64(p.nrShards))); conn != nil {
-			return conn
+			if waiting := conn.Waiting(); waiting < minBusy {
```
Choraden, line 45: Please correct me if I'm wrong, but I'm afraid maxuint32 might be negative on 32bit OS. I would go for uint32 as the type for minBusy and Waiting().

Choraden, line 45: Please correct me if I'm wrong, but I'm afraid maxuint32 might be negative on 32bit OS. I would go for uint32 as the type for minBusy and Waiting().

transport/pool.go
```diff
@@ -34,30 +33,34 @@ func NewConnPool(addr string, connConfig ConnConfig) (*ConnPool, error) {
...
-	return p.NextConn()
-}
+func (p *ConnPool) LeastBusyConn() *Conn {
+	var (
+		leastBusyConn *Conn
+		minBusy       = math.MaxUint32
+	)
 
-func (p *ConnPool) NextConn() *Conn {
 	idx := p.idx.Add(1)
```
mmatczuk, line 42: Use plain loop, atomic index not needed anymore

Makefile
```diff
@@ -17,13 +17,13 @@ test-no-cache:
 integration-test: RUN=Integration
 integration-test:
 ifeq ($(OS),Linux)
-	go test -v -tags integration -run $(RUN) -race ./transport $(ARGS)
+	go test -v -tags integration -run $(RUN) -race -short ./transport $(ARGS)
```
mmatczuk, line 20: Not related, send separate patch.

Makefile
```diff
@@ -17,13 +17,13 @@ test-no-cache:
...
 else ifeq ($(OS),Darwin)
 	@CGO_ENABLED=0 GOOS=linux go test -v -tags integration -c -o ./integration-test.dev ./transport
 	@docker run --name "integration-test" \
 		--network scylla_go_driver_public \
 		-v "$(PWD)/integration-test.dev:/usr/bin/integration-test:ro" \
-		-it --read-only --rm ubuntu integration-test -test.v -test.run $(RUN) $(ARGS)
+		-it --read-only --rm ubuntu integration-test -test.v -test.run $(RUN) -test.short $(ARGS)
```
mmatczuk, line 26: Not related, send separate patch.

transport/conn.go
```diff
@@ -412,8 +412,8 @@ func (c *Conn) sendRequest(req frame.Request, compress, tracing bool) (frame.Res
 	return resp.Response, resp.Err
 }
 
-func (c *Conn) Waiting() int {
-	return int(c.metrics.InQueue.Load() + c.metrics.InFlight.Load())
+func (c *Conn) Waiting() uint32 {
+	return c.metrics.InQueue.Load() + c.metrics.InFlight.Load()
```
mmatczuk, line 416: Why make API ugly?

transport/conn_integration_test.go
```diff
@@ -131,6 +131,10 @@ func BenchmarkConnQueryIntegration(b *testing.B) {
 }
 
 func TestCloseHangingIntegration(t *testing.T) {
+	if testing.Short() {
```
mmatczuk, line 134: WTF?

transport/pool.go
```diff
@@ -34,30 +32,33 @@ func NewConnPool(addr string, connConfig ConnConfig) (*ConnPool, error) {
...
-	conn := p.loadConn(rand.Intn(p.nrShards))
-	if conn != nil {
-		return conn
-	}
-	return p.NextConn()
-}
+func (p *ConnPool) LeastBusyConn() *Conn {
+	var (
+		leastBusyConn *Conn
+		minBusy       = uint32(math.MaxUint32)
```
mmatczuk, line 38: Let's use max stream ID here.

mmatczuk: as int

mmatczuk (changes requested):
Not only this PR contains some unrelated code but it also does not implement what's requested.

Note

```
And there is another connection that's at least 20% less busy. Then we use the least busy connection instead of the one that connects to the right shard.
```

Neofine:
> this PR contains some unrelated code

My apologies, that wasn't intentional, quite frankly I am not sure how this code got there as I haven't made these changes.

> And there is another connection that's at least 20% less busy. Then we use the least busy connection instead of the one that connects to the right shard.

Will do! Thanks!

transport/conn.go
Neofine: Does this make API ugly? ;o
During studies we were taught that greater specification of types lead to more understandable code,
thanks for the tip!

Neofine:
```go

func (c *Conn) moreLoadedThan(leastBusy *Conn) bool {
	return int(float64(c.Waiting())*loadDiffThreshold) > leastBusy.Waiting()
}

func (c *Conn) isHeavyLoaded() bool {
	return int(float64(c.Waiting())/heavyLoadThreshold) > maxStreamID
}
```

This may seem a bit clunky but allows us to have easily modifiable constants.

mmatczuk:
I like the idea.

transport/pool_integration_test.go
```diff
@@ -51,3 +54,59 @@ func TestNodeConnPoolIntegration(t *testing.T) {
 		}
 	}
 }
+
+func TestConnPoolGetConn(t *testing.T) {
```
mmatczuk, line 58: Conn not GetConn

mmatczuk: I think this test is kind of a smoke test and it seems it to me it's not correct.
From the fact that a connection is returned you cannot draw a conclusion that its Waiting value changes.
Please remove the test.

the gist of it can be added to the test that exists.

Also i

mmatczuk (changes requested):
@Neofine remove the test or fix it

I appreciate if you could fix the ordering of functions - sort by dependency most general first.

Conn
helper functions in order of appearance
LeaseBusyConn

transport/pool_integration_test.go
Neofine: > From the fact that a connection is returned you cannot draw a conclusion that its Waiting value changes.

I agree that's why after drawing connection I call `_, _ = conn.Supported()`.
Which calls `c.sendRequest` and it calls `c.r.setHandler(h)` which although may return error and won't change connection Waiting value (it's returning error when `connReader.closed == true` or `c.s.Alloc()` fails when all stream IDs are busy), but in the span of having semi-new connection and 100k requests I think extremely unlikely that it will fail.
If `c.r.setHandler(h)` doesn't fail then connection Waiting value changes.

If I have a flaw in my thought process please correct me and I will incorporate the MurmurToken part to the `TestNodeConnPoolIntegration` and remove the rest.

mmatczuk:
Props for pushing it and fixing the test 💯.

PR #145, opened 2022-03-08
Conn: send and rcv on error should close connection

Neofine:
Idea is based on letting sendRequest() function handle connection closing as it is waiting on a channel and we can use this channel to communicate with it by using errors.

### DIFF
--- a/transport/conn.go
+++ b/transport/conn.go
@@ -49,6 +49,7 @@ type connWriter struct {
 	buf       frame.Buffer
 	requestCh chan request
 	metrics   *connMetrics
+	connClose func()
 }
 
 func (c *connWriter) submit(r request) {
@@ -77,7 +78,10 @@ func (c *connWriter) loop() {
 
 			if err := c.send(r); err != nil {
 				r.ResponseHandler <- response{Err: fmt.Errorf("send: %w", err)}
-				// TODO: notify connection owner that connection needs to be replaced.
+				if c.connClose != nil {
+					c.connClose()
+				}
+				return
 			}
 		}
 		c.metrics.InFlight.Add(uint32(size))
@@ -116,6 +120,7 @@ type connReader struct {
 	bufw        io.Writer
 	metrics     *connMetrics
 	handleEvent func(r response)
+	connClose   func()
 
 	h      map[frame.StreamID]responseHandler
 	s      streamIDAllocator
@@ -171,6 +176,9 @@ func (c *connReader) loop() {
 
 		if resp.Err != nil {
 			c.drainHandlers()
+			if c.connClose != nil {
+				c.connClose()
+			}
 			return
 		}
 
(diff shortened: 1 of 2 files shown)


mmatczuk:
How about closing the connection on the spot?

mmatczuk:
We have put a lot of work to make sure Close is safe, it's connected to pool, let's use it.
Let's also log something.

Neofine:
> We have put a lot of work to make sure Close is safe, it's connected to pool, let's use it.

`Close()` is a method on `Conn` struct and we currently have no access to it from in neither of `ConnWriter` nor `connReader` so we would have to add `*Conn` to their structs or to the argument of their loops which both seem like not desirable solution.

> Let's also log something.

Great idea, thanks!

mmatczuk:
If things go south it should auto close. Pool would replace it.

transport/conn.go
```diff
@@ -424,9 +440,15 @@ func (c *Conn) Shard() int {
 	return int(c.shard)
 }
 
+func (c *Conn) deactivateConnClose() {
+	c.w.connClose = nil
+	c.r.connClose = nil
+}
+
```
mmatczuk, line 447: This is clearly a race.

mmatczuk: I suggest we use [sync.Once](https://pkg.go.dev/sync#example-Once) in Conn::Close to ensure the actual closing happens once.

transport/conn.go
```diff
@@ -325,6 +340,7 @@ func WrapConn(conn net.Conn) (*Conn, error) {
 		},
 		metrics: m,
 	}
+	c.setConnClose()
```
mmatczuk, line 343: Please inline, and remove deactivateConnClose

transport/conn.go
```diff
@@ -116,10 +120,11 @@ type connReader struct {
...
 
-	h      map[frame.StreamID]responseHandler
-	s      streamIDAllocator
-	closed bool
-	mu     sync.Mutex // mu guards h, s and closed
+	h         map[frame.StreamID]responseHandler
+	s         streamIDAllocator
+	closed    bool
+	mu        sync.Mutex // mu guards h, s and closed
+	connClose func()
```
mmatczuk, line 127: Not this group, add that before metrics.

transport/conn.go
```diff
@@ -307,34 +318,42 @@ func OpenConn(addr string, localAddr *net.TCPAddr, cfg ConnConfig) (*Conn, error
...
 	m := new(connMetrics)
-	c := &Conn{
+	var c Conn
+	c = Conn{
 		conn: conn,
 		w: connWriter{
 			conn:      bufio.NewWriterSize(conn, ioBufferSize),
 			requestCh: make(chan request, requestChanSize),
 			metrics:   m,
+			connClose: c.closeOnce,
```
Neofine, line 336: To make this assign possible I had to declare `c` earlier in code that's why I have added
`var c Conn`, is it worth it?
Other option is just keep it as it was and add
```
c.w.connClose = c.connClose
c.r.connClose = c.connClose
```
after the `&Conn` declaration, what do you think?

mmatczuk:
@Neofine I added 44bbb1c8d7e3d3c64d1f02ec88aee400b01f3416 to improve the closing.

Nice trick with var init before use.

Issue #140, opened 2022-03-04
transport: conn c.conn.Flush() error is not checked


mmatczuk:
* Why this is not reported by linter @Choraden 
* Error handling must be added @Neofine

PR #146, opened 2022-03-08
Conn: Added appropriate responses to the recorded errors

Neofine:
- When bufio.writer.Flush() records error it sends to the waiting channels which requests failed to be sent an error response

- Error on conn.Close() logs it

Fixes https://github.com/mmatczuk/scylla-go-driver/issues/140

### DIFF
--- a/transport/conn.go
+++ b/transport/conn.go
@@ -320,12 +320,6 @@ func OpenConn(addr string, localAddr *net.TCPAddr, cfg ConnConfig) (*Conn, error
 	return WrapConn(tcpConn)
 }
 
-func (c *Conn) setWriterReseter() {
-	c.w.reset = func() {
-		c.w.conn.Reset(c.conn)
-	}
-}
-
 // WrapConn transforms tcp connection to a working Scylla connection.
 // If error and connection are returned the connection is not valid and must be closed by the caller.
 func WrapConn(conn net.Conn) (*Conn, error) {
@@ -344,7 +338,10 @@ func WrapConn(conn net.Conn) (*Conn, error) {
 		},
 		metrics: m,
 	}
-	c.setWriterReseter()
+	c.w.reset = func() {
+		c.w.conn.Reset(c.conn)
+	}
+
 	go c.w.loop()
 	go c.r.loop()
 
--- a/transport/conn.go
+++ b/transport/conn.go
@@ -49,6 +49,7 @@ type connWriter struct {
 	buf       frame.Buffer
 	requestCh chan request
 	metrics   *connMetrics
+	reset     func()
 }
 
 func (c *connWriter) submit(r request) {
@@ -62,6 +63,8 @@ func (c *connWriter) loop() {
 	// When requests pile up, allow sending up to 10% in one syscall.
 	var maxCoalescedRequests = requestChanSize / 10
 
+	recentRequests := make([]*request, maxCoalescedRequests)
+
 	for {
 		size := len(c.requestCh)
 		if size > maxCoalescedRequests {
@@ -70,6 +73,7 @@ func (c *connWriter) loop() {
 
 		for i := 0; i < size; i++ {
 			r := <-c.requestCh
+			recentRequests[i] = &r
 			if r == _connCloseRequest {
 				return
 			}
@@ -81,7 +85,16 @@ func (c *connWriter) loop() {
 			}
 		}
 		c.metrics.InFlight.Add(uint32(size))
-		c.conn.Flush()
+
+		if err := c.conn.Flush(); err != nil {
+			for i := 0; i < size; i++ {
+				recentRequests[i].ResponseHandler <- response{Err: fmt.Errorf("send: %w", err)}
+			}
+			if c.reset != nil {
+				c.reset()
+			}
+			// TODO: else close Conn
+		}
 	}
 }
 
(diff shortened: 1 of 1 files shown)


transport/conn.go
```diff
@@ -427,7 +433,12 @@ func (c *Conn) Shard() int {
 // Close closes connection and terminates reader and writer go routines.
 func (c *Conn) Close() {
 	log.Printf("%s closing", c)
-	_ = c.conn.Close()
+	err := c.conn.Close()
+	if err != nil {
+		// Error happens when either conn or file descriptor is nil or there
+		// has been error when closing file descriptor.
+		log.Printf("%s conn.Close() error", c)
```
Neofine, line 440: Should we add some go func() that once in a while tries again to close connections on which conn.Close() errors have been recorded?
Worst case now we have open file descriptor which seems pretty bad.

transport/conn.go
```diff
@@ -81,7 +81,13 @@ func (c *connWriter) loop() {
 			}
 		}
 		c.metrics.InFlight.Add(uint32(size))
-		c.conn.Flush()
+		err := c.conn.Flush()
+		// bufio.Writer doesn't write if error has been recorded, connection needs to be closed.
+		// We can't use writer.Reset() because it would discard any not flushed data.
+		if err != nil {
+			// TODO: notify connection owner that connection needs to be replaced.
+			return
```
Neofine, line 89: I think this seems like pretty reasonable behavior only `writer.Reset()` may discard error on writer but as it deletes any not flushed data it may cause for a thread to hang on `sendRequest()` indefinitely as neither connWriter nor connReader processes its request.

Choraden:
Let's move assignments to if statements. 
if err := ...;...  {}

transport/conn.go
Choraden: Yes, once flush failed connection must be closed.

transport/conn.go
Choraden: I don't think it works like that. 
log.Fatal seems to be reasonable at this point

Neofine: Based on my quick internet sweep I haven't found projects that actually care about error on `conn.Close()` so I think regular `log.Printf()` will be more than enough.

> Should we add some go func() that once in a while tries again to close connections on which conn.Close() errors have been recorded?

This isn't that big of a deal to do that now I guess.

Choraden: agree

transport/conn.go
```diff
@@ -427,7 +433,12 @@ func (c *Conn) Shard() int {
 // Close closes connection and terminates reader and writer go routines.
 func (c *Conn) Close() {
 	log.Printf("%s closing", c)
-	_ = c.conn.Close()
+	err := c.conn.Close()
+	if err != nil {
+		// Error happens when either conn or file descriptor is nil or there
```
mmatczuk, line 438: 1 + 1 = 2

transport/conn.go
mmatczuk, line 440: Why the error is not logged?

mmatczuk, line 440: log.Printf("%s failed to close: %s", c, err)

mmatczuk (changes requested):
This patch completely misses the point...

- On c.conn.Flush() error WE MUST NOTIFY ALL INVOLVED HANDLERS, like in `r.ResponseHandler <- response{Err: fmt.Errorf("send: %w", err)}`
- I do not appreciate the trivial comments, if they help you in coding it's fine, no need to merge them

mmatczuk (changes requested):
Please rebase btw.

Michal-Leszczynski:
What if here:
```
size := len(c.requestCh)
		if size > maxCoalescedRequests {
			size = maxCoalescedRequests
		}
```
**size** is equal to 0?
In that case, don't we have some form of active waiting?
Should we perhaps set size to be greater or equal to 1?

transport/conn.go
```diff
@@ -307,6 +320,14 @@ func OpenConn(addr string, localAddr *net.TCPAddr, cfg ConnConfig) (*Conn, error
 	return WrapConn(tcpConn)
 }
 
+func (c *Conn) resetWriterConn() {
+	c.w.conn.Reset(c.conn)
+}
+
+func (c *Conn) setWriterReseter() {
```
Michal-Leszczynski, line 327: Can't we have only `setWriterReseter` with `resetWriterConn` inlined? Or does the latter one have other uses?

Neofine:
> 



> What if here:
> 
> ```
> size := len(c.requestCh)
> 		if size > maxCoalescedRequests {
> 			size = maxCoalescedRequests
> 		}
> ```
> 
> **size** is equal to 0? In that case, don't we have some form of active waiting? Should we perhaps set size to be greater or equal to 1?

I second that motion. Right now we do have active waiting. 
@mmatczuk shall we deal with this in another issue or should I fix this here?

transport/conn.go
```diff
@@ -325,6 +338,10 @@ func WrapConn(conn net.Conn) (*Conn, error) {
 		},
 		metrics: m,
 	}
+	c.w.reset = func() {
+		c.w.conn.Reset(c.conn)
```
Neofine, line 342: Should I choose here the same approach as in [send rcv error PR](https://github.com/mmatczuk/scylla-go-driver/pull/145#discussion_r828505756)?

mmatczuk:
If size is 0 than you wait on channel it mean there is no request.

mmatczuk:
It looks that I was wrong and plain close on flush error is ok because it will be handled by draining handlers.

PR #148, opened 2022-03-08
Pool: Added changing to given by Scylla shard aware port and increase…

Neofine:
…d wait times on tests

- When given port other than Scylla shard aware port pool now changes it to the right one

- Increased wait times on the tests - it was this bug that made the tests fail


Choraden:
Storing address and port separetly would be more clean imo.

transport/pool.go
```diff
@@ -105,6 +106,16 @@ type PoolRefiller struct {
 	active int
 }
 
+func changePort(addr, port *string) error {
```
mmatczuk, line 109: There is net.SplitHostPort

This does not work with IPv6

transport/pool_integration_test.go
```diff
@@ -14,7 +14,7 @@ func TestNodeConnPoolIntegration(t *testing.T) {
 	}
 
 	t.Log("Wait for refiller to fill connections to shards")
-	time.Sleep(250 * time.Millisecond)
+	time.Sleep(500 * time.Millisecond)
```
mmatczuk, line 17: Put that to a variable.

transport/pool.go
```diff
@@ -118,6 +129,13 @@ func (r *PoolRefiller) initConnPool() error {
 	}
 	ss := s.ScyllaSupported()
 
+	if s, ok := s.Options["SCYLLA_SHARD_AWARE_PORT"]; ok && !strings.HasSuffix(r.addr, ":"+s[0]) {
+		err = changePort(&r.addr, &s[0])
+		if err != nil {
+			return fmt.Errorf("port change: %w", err)
+		}
+	}
+
```
mmatczuk, line 138: This is very puzzling change.

I see you try to start with Shard Aware port, it's not needed IMO.

transport/pool.go
```diff
@@ -118,6 +129,13 @@ func (r *PoolRefiller) initConnPool() error {
 	}
 	ss := s.ScyllaSupported()
 
+	if s, ok := s.Options["SCYLLA_SHARD_AWARE_PORT"]; ok && !strings.HasSuffix(r.addr, ":"+s[0]) {
```
Michal-Leszczynski, line 132: I would consider renaming variable s since it has already been used (with different meaning) inside this scope.

Moreover, we have const defined for string "SCYLLA_SHARD_AWARE_PORT" inside response/supported.go.
I would consider using it.

transport/pool.go
Michal-Leszczynski, line 109: I would consider getting rid of pointers (it is not clear to me why are they needed here) and just return new string alongside with potential error.

transport/pool.go
Michal-Leszczynski: So can we assume (for now) that we are given shard aware ports while creating session?
For example, TestNodeConnPoolIntegration won't pass if we create pool with port 9042 instead of 19042
(we won't always be able to create connections to every shard since it does not depend on us in this scenario).

In general it is quiet easy to get shard aware port after connecting to a node (in contrary to getting any port from a node that we are not connected to yet) so I would consider updating shard aware port.

mmatczuk:
You opened a connection, connection is to shard X you know that from supported, what it the problem????

Michal-Leszczynski:
> You opened a connection, connection is to shard X you know that from supported, what it the problem????

Without shard aware port we cannot be sure that we will be able to cover all shards with connections (we could be opening connections to the same shard multiple times). That's why TestConnPoolIntegration sometimes didn't pass, when we were establishing pool with port 9042 instead of 19042.

First connection can be opened on regular port and be mapped to whatever shard, but wouldn't it be nice if we could ensure that other connections are mapped as we please? I would consider updating shard aware port, since it is easy and we already have the data in scylla supported struct.

mmatczuk:
You can assume / hard code as constants both ports.
You then open the 1st conn to 9042 and all the others to the shared aware port.

mmatczuk:
You can then assume the host passed to the pool has no port.

mmatczuk:
It can be later validated on the config level.

mmatczuk (changes requested):
After closer investigation the patch idea IS correct.

- Refiller shall use and save ShardAware port from `ss := s.ScyllaSupported()`
- The conn address should be decorated (not overwritten)
- The decorating function takes address and port and returns address with set or replaced port
- net package has all the manipulation functions
- test for the decorating function is more than wellcome
- in case of error in address manipulation we can log FATAL resulting in exit.

mmatczuk:
Maybe panic is better than log fatal.

mmatczuk:
All tests in X_integration_test.go should be suffixed with `Integration`.
Not integration tests shall be put to X_test.

mmatczuk:
Why we have here handling of `[ADDR]` ?

mmatczuk:
@Neofine please take a look at the alternate implementation I just pushed.

mmatczuk (changes requested):
.

Neofine:
Thanks for the commit, it's much cleaner version of mine, sorry, I got confused by the `The conn address should be decorated (not overwritten)` comment and tried my best to apply it to mine version.
I have commited some tweaks to the `withDefaultPort(addr string)` and `withPort(addr, port string)` functions, mainly added error checking. 
Error may happen if address has no port with it or it has some mistakes, either way I call `net.JoinHostPort` on it, dialing in `OpenConn` will verify its validity.
I had to add 
```go
func trimIPv6Brackets(host string) string {
	host = strings.TrimPrefix(host, "[")
	return strings.TrimSuffix(host, "]")
}

```
function. If error (no port) is detected in IPv6 address we are left with either address in square brackets  or not, so just in case we have to trim them not to create double square brackets with which `net` functions can't deal properly. I can't seem to find cleaner way of doing it, any suggestions are welcome.
I have also changed tests names and created `pool_test.go` to test those port functions.

mmatczuk:
Why IPb6 addresses have this `[]` this is not part of the [spec](https://en.wikipedia.org/wiki/IPv6_address).
Is this is something added by Scylla?
If so it should be taken care of in cluster @Michal-Leszczynski.

PR #159, opened 2022-03-16
Transport: Pool, Conn: Added keyspace to conn and refiller

### LINKED ISSUE #153: conn: add UseKeyspace

Neofine:
- Added "IO" prefix to errors got by sendRequest caused by connection/internal error to differentiate them from other errors
- UseKeyspace can be called to try change keyspace of every present and future connection in connection pool

Fixes #153

### DIFF
--- a/transport/conn.go
+++ b/transport/conn.go
@@ -11,6 +11,7 @@ import (
 	"strconv"
 	"sync"
 	"time"
+	"unicode"
 
 	"scylla-go-driver/frame"
 	. "scylla-go-driver/frame/request"
@@ -266,6 +267,7 @@ type Conn struct {
 }
 
 type ConnConfig struct {
+	Keyspace           string
 	TCPNoDelay         bool
 	Timeout            time.Duration
 	DefaultConsistency frame.Consistency
@@ -327,12 +329,12 @@ func OpenConn(addr string, localAddr *net.TCPAddr, cfg ConnConfig) (*Conn, error
 		return nil, fmt.Errorf("set TCP no delay option: %w", err)
 	}
 
-	return WrapConn(tcpConn)
+	return WrapConn(tcpConn, cfg)
 }
 
 // WrapConn transforms tcp connection to a working Scylla connection.
 // If error and connection are returned the connection is not valid and must be closed by the caller.
-func WrapConn(conn net.Conn) (*Conn, error) {
+func WrapConn(conn net.Conn, cfg ConnConfig) (*Conn, error) {
 	m := new(connMetrics)
 
 	c := new(Conn)
@@ -360,9 +362,40 @@ func WrapConn(conn net.Conn) (*Conn, error) {
 		return c, err
 	}
 
+	if cfg.Keyspace != "" {
+		if err := c.UseKeyspace(cfg.Keyspace); err != nil {
+			return c, fmt.Errorf("use keyspace %w", err)
+		}
+	}
+
 	log.Printf("%s connected", c)
 
 	return c, nil
+}
+
+func (cfg *ConnConfig) validate() error {
+	if cfg.Keyspace != "" {
+		if err := validateKeyspace(cfg.Keyspace); err != nil {
+			return err
+		}
+	}
+	if cfg.DefaultConsistency < frame.ANY || cfg.DefaultConsistency > frame.LOCALONE {
+		return fmt.Errorf("unknown consistency: %v", cfg.DefaultConsistency)
+	}
+	return nil
+}
+
+func validateKeyspace(keyspace string) error {
+	if keyspace == "" || len(keyspace) > 48 {
+		return fmt.Errorf("keyspace: invalid length")
+	}
+
+	for _, c := range keyspace {
+		if !(unicode.IsLetter(c) || unicode.IsDigit(c) || c == '_') {
+			return fmt.Errorf("keyspace: illegal characters present")
+		}
+	}
+	return nil
 }
 
 var startupOptions = frame.StartupOptions{"CQL_VERSION": "3.0.0"}
@@ -399,6 +432,11 @@ func (c *Conn) Startup(options frame.StartupOptions) error {
 		return nil
 	}
 	return responseAsError(res)
+}
+
+func (c *Conn) UseKeyspace(ks string) error {
+	_, err := c.Query(newStatementFromCQL(fmt.Sprintf("USE %q", ks)), nil)
+	return err
 }
 
 func (c *Conn) Query(s Statement, pagingState frame.Bytes) (QueryResult, error) {
--- a/transport/pool.go
+++ b/transport/pool.go
@@ -120,8 +120,15 @@ type PoolRefiller struct {
 }
 
 func (r *PoolRefiller) init(addr string) error {
+	if err := r.cfg.validate(); err != nil {
+		return fmt.Errorf("config validate :%w", err)
+	}
+
 	conn, err := OpenConn(withDefaultPort(addr), nil, r.cfg)
 	if err != nil {
+		if conn != nil {
+			conn.Close()
+		}
 		return err
 	}
 
@@ -202,6 +209,9 @@ func (r *PoolRefiller) fill() {
 		conn, err := OpenShardConn(r.addr, si, r.cfg)
 		if err != nil {
 			log.Printf("failed to open shard conn: %s", err)
+			if conn != nil {
+				conn.Close()
+			}
 			continue
 		}
 		if conn.Shard() != i {
--- a/transport/pool_integration_test.go
+++ b/transport/pool_integration_test.go
@@ -10,8 +10,8 @@ import (
 
 const refillerBackoff = 500 * time.Millisecond
 
-func TestConnPoolIntegration(t *testing.T) {
-	p, err := NewConnPool(TestHost, ConnConfig{})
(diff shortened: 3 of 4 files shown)


mmatczuk:
In future try to put the additional stuff like "Added "IO" prefix to errors got by sendRequest " to separate commits.

transport/pool_integration_test.go
```diff
@@ -10,7 +10,7 @@ import (
 
 const refillerBackoff = 250 * time.Millisecond
 
-func TestConnPoolIntegration(t *testing.T) {
```
mmatczuk, line 13: Naming can be improved.
What you have here is more often called newTestConnPool.

Typically a test helper it is a struct of its own, for example containing a pool as a field - not needed here IMO.

transport/pool_integration_test.go
```diff
@@ -88,4 +80,52 @@ func TestConnPoolConnIntegration(t *testing.T) {
 	if conn := p.Conn(t1); conn == nil {
 		t.Fatal("invalid return of Conn")
 	}
+
+	p.Close()
+}
+
+func TestConnPoolKeyspaceIntegration(t *testing.T) {
+	p := newConnPoolTestHelper(t)
```
mmatczuk, line 88: Use [defer](https://gobyexample.com/defer) for closing, it's more safe

transport/conn.go
```diff
@@ -454,6 +455,16 @@ func (c *Conn) sendRequest(req frame.Request, compress, tracing bool) (frame.Res
 	return resp.Response, resp.Err
 }
 
+func (c *Conn) useKeyspace(keyspace string) error {
```
mmatczuk, line 458: Needs to be Exported.

transport/conn.go
```diff
@@ -454,6 +455,16 @@ func (c *Conn) sendRequest(req frame.Request, compress, tracing bool) (frame.Res
...
+	s := &Statement{Content: `USE "` + keyspace + `"`}
```
mmatczuk, line 459: We could have a constructor `newSatementFromCQL(col sting)` the help with that.
Also applies to cluster and control conn @Michal-Leszczynski.

mmatczuk, line 459: There is a trick you could use `fmt.Sprintf("USE %q", ks)`.

transport/conn.go
```diff
@@ -454,6 +455,16 @@ func (c *Conn) sendRequest(req frame.Request, compress, tracing bool) (frame.Res
...
+	_, err := c.Query(*s, nil)
+	if err != nil {
+		return err
+	}
+	c.keyspace = keyspace
```
mmatczuk, line 464: Unprotected access to a variable - aka race.
I think that technically there is no race here due to linearization in pool.

Let's remove the keyspace from Conn.

transport/conn.go
mmatczuk: I'd suggest moving it after Startup, it's often the 1st thing you want to do post Startup.

transport/pool.go
```diff
@@ -35,6 +46,39 @@ func NewConnPool(addr string, cfg ConnConfig) (*ConnPool, error) {
...
+	}
+	resp := make(keyspaceRespCh)
+	p.keyspaceReqCh <- keyspaceReq{
+		keyspace: keyspace,
+		resp:     resp,
+	}
+	return <-resp
+}
+
+func validKeyspace(keyspace string) error {
```
mmatczuk, line 61: s/valid/validate/ 

Valid more isValid would return bool, typically we call that validateX

transport/pool.go
```diff
@@ -5,18 +5,29 @@ import (
...
+	"strings"
 	"time"
+	"unicode"
 
 	"go.uber.org/atomic"
 )
 
 var _poolCloseShard = -1
 
+type keyspaceRespCh chan error
```
mmatczuk, line 17: This type def has a smell of overengineering, let's inline that.

mmatczuk:
I have a radical idea, add keyspace to NewConnPool.
If empty skip the setting.
If not call UseKeyspace on every new connection.

There are two things that we need to take into consideration.
- the keyspace in never changed on the fly, an app operates in a keyspace period.
- if you manage to change the keyspace in some but not all connections you are screwed

Overall I see coding skill improvements, keep it up @Neofine

transport/pool.go
```diff
@@ -119,12 +121,25 @@ type PoolRefiller struct {
...
 
+	if ks != "" {
+		if err := validateKeyspace(ks); err != nil {
+			return err
+		}
+		if err := conn.UseKeyspace(ks); err != nil {
+			log.Printf("failed to connect to the keyspace: %s", err)
+			conn.Close()
+			if !isIOError(err) {
+				return fmt.Errorf("keyspace request failed %w", err)
```
Neofine, line 138: Should we fail initialization now also when IO error happens?

transport/conn.go
```diff
@@ -417,6 +422,14 @@ func (c *Conn) QueryAsync(s Statement, pagingState frame.Bytes, fn func(QueryRes
 	}()
 }
 
+func (c *Conn) newStatementFromCQL(cql string) (QueryResult, error) {
```
mmatczuk, line 425: Nah, this should make a Statement and it's not a Conn method.

Making a shortcut like this cheats on code reusability - that is future you!

transport/pool.go
```diff
@@ -20,13 +21,14 @@ type ConnPool struct {
 	msbIgnore    uint8
 	conns        []atomic.Value
 	connClosedCh chan int // notification channel for when connection is closed
+	keyspace     string
 }
 
-func NewConnPool(addr string, cfg ConnConfig) (*ConnPool, error) {
+func NewConnPool(addr, ks string, cfg ConnConfig) (*ConnPool, error) {
```
mmatczuk, line 27: Can we name it keyspace here for clarity?

transport/pool.go
mmatczuk: We fail on any error.

We would handle that later with retries - no need to worry now.

transport/pool.go
```diff
@@ -160,6 +176,23 @@ func (r *PoolRefiller) onConnClose(conn *Conn) {
...
+			return fmt.Errorf("keyspace: illegal characters present")
+		}
+	}
+	return nil
+}
+
+func isIOError(err error) bool {
+	return strings.HasPrefix(err.Error(), "IO")
+}
+
```
mmatczuk, line 195: Ha Ha Ha

https://pkg.go.dev/errors - error wrapping and unwrapping
https://cs.opensource.google/go/go/+/refs/tags/go1.18:src/net/net.go;l=439 - the I/O errors, look for error types

transport/conn.go
```diff
@@ -74,7 +74,7 @@ func (c *connWriter) loop() {
 			}
 			if err := c.send(r); err != nil {
 				log.Printf("fatal send error, closing connection due to %s", err)
-				r.ResponseHandler <- response{Err: fmt.Errorf("send: %w", err)}
+				r.ResponseHandler <- response{Err: fmt.Errorf("send: %w [IO]", err)}
```
mmatczuk, line 77: Please do not add the `[IO]` thing.

transport/conn.go
```diff
@@ -197,7 +197,7 @@ func (c *connReader) drainHandlers() {
 	c.mu.Lock()
 	c.closed = true
 	for _, h := range c.h {
-		h <- response{Err: fmt.Errorf("connection closed")}
+		h <- response{Err: fmt.Errorf("connection closed [IO]")}
```
mmatczuk, line 200: There is really no reason to add it.

transport/pool.go
```diff
@@ -119,12 +122,22 @@ type PoolRefiller struct {
...
 		return err
 	}
 
+	if keyspace != "" {
+		if err := validateKeyspace(keyspace); err != nil {
+			return err
+		}
+		if err := conn.UseKeyspace(keyspace); err != nil {
+			conn.Close()
+			return fmt.Errorf("keyspace request failed %w", err)
```
mmatczuk, line 137: `use keyspace` would be easier to understand by a user.

Neofine:
I have removed IO errors from connection as those are not used now.

transport/conn.go
```diff
@@ -269,6 +269,7 @@ type ConnConfig struct {
 	TCPNoDelay         bool
 	Timeout            time.Duration
 	DefaultConsistency frame.Consistency
+	Keyspace           string
```
mmatczuk, line 272: Probably this should go 1st as the most commonly set.

transport/pool.go
```diff
@@ -20,13 +21,14 @@ type ConnPool struct {
 	msbIgnore    uint8
 	conns        []atomic.Value
 	connClosedCh chan int // notification channel for when connection is closed
+	keyspace     string
```
mmatczuk, line 24: I think we do not need this anymore.

transport/pool.go
```diff
@@ -20,13 +21,14 @@ type ConnPool struct {
...
 }
 
 func NewConnPool(addr string, cfg ConnConfig) (*ConnPool, error) {
 	r := PoolRefiller{
 		cfg: cfg,
 	}
-	if err := r.init(addr); err != nil {
+	if err := r.init(addr, cfg.Keyspace); err != nil {
```
mmatczuk, line 31: Same here.

mmatczuk:
We need to remove more code in the pool, the validation could go to ConnConfig Validate function.

PR #168, opened 2022-03-21
frame: cqlvalue, added AsStringSlice()

Kulezi:
@Michal-Leszczynski mentioned the need for parsing lists of token strings in cluster, those are actually of `set<text>` type.
This PR adds CqlValue.AsTextSet() allowing parsing token sets.

Also I think that validating correctness of CqlValue underlying bytes (e.g. checking if strings are utf8) should be done only when Debug flag is set, as we shouldn't get invalid data from queries. Correct me if I'm wrong.

### DIFF
--- a/frame/cqlvalue.go
+++ b/frame/cqlvalue.go
@@ -1,6 +1,7 @@
 package frame
 
 import (
+	"encoding/binary"
 	"fmt"
 	"math"
 	"net"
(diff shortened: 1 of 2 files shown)


mmatczuk:
Naming, can we use String instead of Text and Slice instead of Set i.e. []string is StringSlice.
This is a general comment applies to some existing code too.

mmatczuk:
Also, technically CqlFromTextSet creates a List?
A more flexible naming convention is needed.

frame/cqlvalue.go
```diff
@@ -198,6 +199,31 @@ func (c CqlValue) AsFloat64() (float64, error) {
 		uint64(c.Value[7])), nil
 }
 
+func (c CqlValue) AsTextSet() ([]string, error) {
+	if c.Type.ID != SetID || c.Type.List.Element.ID != VarcharID {
```
Michal-Leszczynski, line 203: I get errors when using this function. There should be:
`c.Type.Set.Element.ID`
instead of:
`c.Type.List.Element.ID`

Michal-Leszczynski (changes requested):

Kulezi:
Renamed to AsStringSlice(), this feels more natural, it allows parsing from lists and sets of varchar, text and ascii types.

Choraden (approved):

Michal-Leszczynski (approved):

Kulezi:
Merged with #179

Issue #116, opened 2022-01-28
TestConnMassiveQueryIntegration does not work for large n


mmatczuk:
At the moment the test drops large amounts of queries because Scylla detects the same query repeated over and over.

```
Jan 28 14:26:56 d0f3d59d1516 scylla:  [shard 0] reader_concurrency_semaphore - (rate limiting dropped 1363 similar messages) Semaphore _read_concurrency_sem with 75/100 count and 1246360/10737418 memory resources: timed out, dumping permit diagnostics:#012permits#011count#011memory#011table/description/state#01274#01174#0111184K#011mykeyspace.users/shard-reader/inactive#012272#0110#01117K#011mykeyspace.users/multishard-mutation-query/active/unused#0121#0111#01116K#011mykeyspace.users/shard-reader/active/used#0127#0110#0110B#011mykeyspace.users/shard-reader/waiting#012#012354#01175#0111217K#011total#012#012Total: 354 permits with 75 count and 1217K memory resources
```

PR #173, opened 2022-03-25
transport: conn, add QueryAsync

mmatczuk:
```
BenchmarkConnQueryIntegration-8             2233           2654839 ns/op            1251 B/op         34 allocs/op
BenchmarkConnAsyncQueryIntegration-8       10000           2088146 ns/op            1289 B/op         11 allocs/op
```

- QueryAsync runs Query in a separated goroutine
- Benchmarks moved to separate file
- connTestHelper added applyFixture
- requestChanSize increased to maxStreamID/2, with 1k goroutines block on putting to the channel
- newQueryForStatement changed to makeQueryForStatement returning Query this removes the gc pressure

### DIFF
--- a/transport/conn_integration_bench_test.go
+++ b/transport/conn_integration_bench_test.go
@@ -1,0 +1,55 @@
+//go:build integration
+
+package transport
+
+import (
+	"sync"
+	"testing"
+
+	"scylla-go-driver/frame"
+)
+
+var benchmarkConnQueryResult QueryResult
+
+func BenchmarkConnQueryIntegration(b *testing.B) {
+	h := newConnTestHelper(b)
+	h.applyFixture()
+	defer h.close()
+
+	query := Statement{Content: "SELECT * FROM mykeyspace.users", Consistency: frame.ONE}
+
+	var (
+		r   QueryResult
+		err error
+	)
+	b.ResetTimer()
+	for n := 0; n < b.N; n++ {
+		r, err = h.conn.Query(query, nil)
+		if err != nil {
+			b.Fatal(err)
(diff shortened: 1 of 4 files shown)


transport/query.go
```diff
@@ -16,8 +16,8 @@ type Statement struct {
 	Compression       bool
 }
 
-func newQueryForStatement(s Statement, pagingState frame.Bytes) frame.Request {
-	return &Query{
+func makeQueryForStatement(s Statement, pagingState frame.Bytes) Query {
```
mmatczuk, line 19: @Choraden @Kulezi the same needs to be done for execute and maybe prepare.

In the context of Conn::Query there is an optiomisation for stack pointers, it's better to copy Query to stack and point to stack than allocate it on heap - no GC...

mmatczuk:
Probably the results would be even better but on some requests we face t/o due to semaphore in Scylla https://github.com/mmatczuk/scylla-go-driver/issues/116.

transport/conn_integration_bench_test.go
```diff
@@ -0,0 +1,55 @@
...
+		wg sync.WaitGroup
+		fn = func(fr QueryResult, err error) {
+			wg.Done()
+		}
+	)
+
+	b.ResetTimer()
+	wg.Add(b.N)
+	for n := 0; n < b.N; n++ {
+		h.conn.QueryAsync(query, nil, fn)
```
Choraden, line 52: How are we going to collect results? I think we should add collecting in benchmark, since in classic query we have results in "our hands", and here we still have to somehow grab it.

PR #174, opened 2022-03-25
Ml/host selection policy

### LINKED ISSUE #152: Implement host selection policy
Let's follow Rust transport/load_balancing. 
You can take a look at TokenAwareHostPolicy in gocql but it's a mess.

Michal-Leszczynski:
Fixes #152

### DIFF
--- a/transport/cluster.go
+++ b/transport/cluster.go
@@ -3,22 +3,29 @@ package transport
 import (
 	"fmt"
 	"log"
+	"strconv"
 	"time"
 
 	"scylla-go-driver/frame"
 	. "scylla-go-driver/frame/response"
 
+	"github.com/google/btree"
 	"go.uber.org/atomic"
 )
 
 type (
-	PeerMap = map[string]*Node
+	peerMap    = map[string]*Node
+	dcRacksMap = map[string]int
+	dcRFMap    = map[string]uint32
+	ksMap      = map[string]keyspace
 
 	requestChan chan struct{}
 )
 
+const BTreeDegree = 2
+
 type Cluster struct {
-	peers             atomic.Value // PeerMap
+	topology          atomic.Value // *topology
 	control           *Conn
 	cfg               ConnConfig
 	handledEvents     []frame.EventType // This will probably be moved to config.
@@ -26,16 +33,98 @@ type Cluster struct {
 	refreshChan       requestChan
 	reopenControlChan requestChan
 	closeChan         requestChan
-
-	// Yet to be used.
-	// ring        map[Token]*Node
-	// dataCenters map[string]*DataCenter
 }
 
-// type dataCenter struct {
-// 	nodes   []Node
-// 	rackCnt uint // Type?
-// }
+type topology struct {
+	peers     peerMap
+	dcRacks   dcRacksMap
+	nodes     []*Node
+	ring      *btree.BTree
+	keyspaces ksMap
+}
+
+type keyspace struct {
+	strategy strategy
+	// TODO: Add and use attributes below.
+	// tables		      map[string]table
+	// user_defined_types map[string](string, cqltype)
+}
+
+type strategyClass string
+
+// JCN stands for Java Class Name.
+const (
+	networkTopologyStrategyJCN strategyClass = "org.apache.cassandra.locator.NetworkTopologyStrategy"
+	simpleStrategyJCN          strategyClass = "org.apache.cassandra.locator.SimpleStrategy"
+	localStrategyJCN           strategyClass = "org.apache.cassandra.locator.LocalStrategy"
+	networkTopologyStrategy    strategyClass = "NetworkTopologyStrategy"
+	simpleStrategy             strategyClass = "SimpleStrategy"
+	localStrategy              strategyClass = "LocalStrategy"
+)
+
+type strategy struct {
+	class strategyClass
+	rf    uint32            // Used in simpleStrategy.
+	dcRF  dcRFMap           // Used in networkTopologyStrategy.
+	data  map[string]string // Used in other strategy.
+
+}
+
+// QueryInfo represents data required for host selection policy to create query plan.
+// Token and strategy are only necessary for token aware policies.
+type QueryInfo struct {
+	tokenAwareness bool
+	token          Token
+	topology       *topology
+	strategy       strategy
+}
+
+func (c *Cluster) newQueryInfo() QueryInfo {
+	return QueryInfo{
+		tokenAwareness: false,
+		topology:       c.Topology(),
+	}
+}
+
+func (c *Cluster) newTokenAwareQueryInfo(t Token, ks string) (QueryInfo, error) {
+	top := c.Topology()
+	// When keyspace is not specified, we take default keyspace from ConnConfig.
+	if ks == "" {
+		ks = c.cfg.Keyspace
+	}
+	if stg, ok := top.keyspaces[ks]; ok {
+		return QueryInfo{
+			tokenAwareness: true,
+			token:          t,
+			topology:       top,
+			strategy:       stg.strategy,
+		}, nil
+	} else {
+		return QueryInfo{}, fmt.Errorf("couldn't find given keyspace in current topology")
+	}
+}
+
+// replicas return slice of nodes (desirably of length cnt) holding data described by token.
+// filter function allows applying additional requirements for nodes to be taken.
+func (t *topology) replicas(token Token, size int, filter func(*Node, []*Node) bool) []*Node {
+	res := make([]*Node, 0, size)
+	it := func(i btree.Item) bool {
+		n := i.(RingEntry).node
+		if filter(n, res) {
+			res = append(res, n)
+		}
+		return len(res) < size
+	}
+
+	re := RingEntry{token: token}
+	// Token ring has cyclic architecture, so we also have to
+	// get back to the beginning after reaching its end.
+	t.ring.AscendGreaterOrEqual(re, it)
+	if len(res) < size {
+		t.ring.AscendLessThan(re, it)
+	}
+	return res
+}
 
 // NewCluster also creates control connection and starts handling events and refreshing topology.
 func NewCluster(cfg ConnConfig, e []frame.EventType, hosts ...string) (*Cluster, error) {
@@ -54,7 +143,7 @@ func NewCluster(cfg ConnConfig, e []frame.EventType, hosts ...string) (*Cluster,
 		reopenControlChan: make(requestChan, 1),
 		closeChan:         make(requestChan, 1),
 	}
-	c.setPeers(PeerMap{})
+	c.setTopology(&topology{})
 
 	if control, err := c.NewControl(); err != nil {
 		return nil, fmt.Errorf("create control connection: %w", err)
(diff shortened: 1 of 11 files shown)


Neofine (approved):
Looks good, the only thing we need to pay a closer look to is parallel tests ;)

mmatczuk:
Please do not use parallel test here.

Michal-Leszczynski:
Do we support situations in which currently used keyspace is not defined?

transport/cluster.go
```diff
@@ -3,42 +3,113 @@ package transport
...
 	"scylla-go-driver/frame"
 	. "scylla-go-driver/frame/response"
 
+	"github.com/google/btree"
 	"go.uber.org/atomic"
 )
 
 type (
-	PeerMap = map[string]*Node
+	PeerMap         = map[string]*Node
```
mmatczuk, line 17: Do we need to export all those helper types?

transport/cluster.go
```diff
@@ -3,42 +3,113 @@ package transport
...
+	DCRacksMap      = map[string]int
+	DCRepFactorsMap = map[string]repFactor
+	KSMap           = map[string]keyspace
```
mmatczuk, line 20: Do we need to export all those helper types?

transport/cluster.go
```diff
@@ -3,42 +3,113 @@ package transport
...
 
 	requestChan chan struct{}
+
+	repFactor uint32
```
mmatczuk, line 24: Often referred as `rf`, that is to say dcRfMap looks better. 
Not sure we need a type for that.
This is an int a small one usually `3`, what is the benefit of the dedicated type?

transport/cluster.go
```diff
@@ -3,42 +3,113 @@ package transport
...
+	// TODO: Add and use attributes below
+	// tables		      map[string]table
+	// user_defined_types map[string](string, cqltype)
 }
 
-// type dataCenter struct {
-// 	nodes   []Node
-// 	rackCnt uint // Type?
-// }
+type stgClass int
```
mmatczuk, line 56: How about strategyClass?

transport/cluster.go
```diff
@@ -3,42 +3,113 @@ package transport
...
+
+const (
+	networkTopology stgClass = iota
+	simple
+	local
+	other
```
mmatczuk, line 62: Best would be to make it a string and use the java class name as value.

transport/cluster.go
```diff
@@ -3,42 +3,113 @@ package transport
...
+	dcRF DCRepFactorsMap
+	// used in other strategy
+	name string
+	data map[string]string
+}
+
+type QueryInfo struct {
+	token    *Token
+	topology *topology
+	stg      strategy
```
mmatczuk, line 79: I'd prefer strategy vs stg, stg is a good name for a short lived local variable

transport/cluster.go
```diff
@@ -3,42 +3,113 @@ package transport
...
+}
+
+func (c *Cluster) newQueryInfo(t *Token) QueryInfo {
+	top := c.topology.Load().(*topology)
```
mmatczuk, line 83: You can wrap it in some nice function.

transport/cluster.go
```diff
@@ -3,42 +3,113 @@ package transport
...
+	return QueryInfo{
+		token:    t,
+		topology: top,
+		stg:      top.keyspaces[c.keyspace].stg, // TODO: do we assume, that keyspace is always specified?
```
mmatczuk, line 87: If keyspace is not specified the token is nil. Unless you can tell the keyspace from the prepare result.

mmatczuk: Namely https://github.com/apache/cassandra/blob/d70f6e0d597e888b3629c6c01d05685cf79a3d12/doc/native_protocol_v4.spec#L598

transport/cluster.go
```diff
@@ -3,42 +3,113 @@ package transport
...
+	rf repFactor
+	// used in networkTopology strategy
+	dcRF DCRepFactorsMap
+	// used in other strategy
+	name string
+	data map[string]string
+}
+
+type QueryInfo struct {
+	token    *Token
```
mmatczuk, line 77:
```
// Token is used to identify both nodes and partitions, it's value is hashed partition key.
type Token struct {
	value int64
}
```

I'd prefer to use `type Token int64`

and use here the value not the pointer.

mmatczuk: All the `t *Token` should be `t Token`.

transport/cluster.go
```diff
@@ -3,42 +3,113 @@ package transport
...
 	refreshChan       requestChan
 	reopenControlChan requestChan
 	closeChan         requestChan
+}
+
+type topology struct {
+	peers     PeerMap
+	dcRacks   DCRacksMap
+	allNodes  []*Node
+	ring      btree.BTree
```
mmatczuk, line 45: Since we have `ring := *btree.New(BTreeDegree)` let's use a pointer here.

transport/cluster.go
```diff
@@ -3,42 +3,113 @@ package transport
...
+	keyspaces KSMap
+}
 
-	// Yet to be used.
-	// ring        map[Token]*Node
-	// dataCenters map[string]*DataCenter
+type keyspace struct {
+	stg strategy
```
mmatczuk, line 50: Full name please.

transport/cluster.go
```diff
@@ -3,42 +3,113 @@ package transport
...
+	}
+}
+
+// ringRange return slice of nodes (desirably of length cnt) holding data described by token.
+// wanted function allows applying additional requirements for nodes to be taken.
+func (t *topology) ringRange(token Token, cnt int, wanted func(*Node, []*Node) bool) []*Node {
+	replicas := make([]*Node, 0, cnt)
+	it := func(i btree.Item) bool {
+		n := i.(RingEntry).node
+		if wanted(n, replicas) {
```
mmatczuk, line 97: why would we pass replicas here?

transport/cluster.go
```diff
@@ -3,42 +3,113 @@ package transport
...
+	return QueryInfo{
+		token:    t,
+		topology: top,
+		stg:      top.keyspaces[c.keyspace].stg, // TODO: do we assume, that keyspace is always specified?
+	}
+}
+
+// ringRange return slice of nodes (desirably of length cnt) holding data described by token.
+// wanted function allows applying additional requirements for nodes to be taken.
+func (t *topology) ringRange(token Token, cnt int, wanted func(*Node, []*Node) bool) []*Node {
```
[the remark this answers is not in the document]
mmatczuk: Shouldn't ringRange be called replicas, returns replicas of a given token?

transport/cluster.go
```diff
@@ -3,42 +3,113 @@ package transport
...
+			replicas = append(replicas, n)
+		}
+		return len(replicas) < cnt
+	}
+
+	re := RingEntry{token: token}
+	t.ring.AscendGreaterOrEqual(re, it)
+	if len(replicas) < cnt {
+		t.ring.AscendLessThan(re, it)
+	}
```
mmatczuk, line 107: Can you give a comment here on what is going on?

transport/cluster.go
```diff
@@ -3,42 +3,113 @@ package transport
...
+	return replicas
+}
 
 // NewCluster also creates control connection and starts handling events and refreshing topology.
-func NewCluster(cfg ConnConfig, e []frame.EventType, hosts ...string) (*Cluster, error) {
+func NewCluster(cfg ConnConfig, ks string, e []frame.EventType, hosts ...string) (*Cluster, error) {
```
mmatczuk, line 112: I'd suggest we put keyspace to ConnConfig

mmatczuk: All the use keyspace stuff can be pushed down to WrapConn and init @Neofine.

transport/cluster.go
```diff
@@ -100,9 +172,18 @@ func (c *Cluster) refreshTopology() error {
...
 	old := c.Peers()
-	m := PeerMap{}
-	for _, v := range rows {
-		n, err := cqlRowIntoNode(v)
+	peers := make(PeerMap)
+	dcRacks := make(DCRacksMap)
+	seenRacks := make(map[string]bool) // Key is rack name appended to its dc name.
+	all := make([]*Node, 0, len(old))
+	ring := *btree.New(BTreeDegree)
+	keyspaces, err := c.parseKeyspace()
```
mmatczuk, line 180: Maybe allocate topology struct pointer instead of doing

```
topology{
		peers:     peers,
		dcRacks:   dcRacks,
		allNodes:  all,
		ring:      ring,
		keyspaces: keyspaces,
	}
```
later?

transport/cluster.go
```diff
@@ -118,18 +199,65 @@ func (c *Cluster) refreshTopology() error {
...
+		ring:      ring,
+		keyspaces: keyspaces,
+	})
+	// Discards request that came to us while refreshing.
+	select {
+	case <-c.refreshChan:
+	default:
+	}
 	return nil
 }
```
mmatczuk, line 239: I'm not quite sure why we need this?

Refactor to method and put close to where the put is.
In general this should be done in a for loop.

transport/cluster.go
```diff
@@ -164,6 +297,12 @@ const (
 	nodeTokens = 3
 )
 
+// Indexes of columns returned from keyspace query.
+const (
+	ksName      = 0
```
mmatczuk, line 302: Worth moving inside function where it's used.

transport/cluster.go
```diff
@@ -177,21 +316,89 @@ func cqlRowIntoNode(r frame.Row) (*Node, error) {
...
-	tokens := r[nodeTokens] // TODO: add parsing for string list.
 	return &Node{
 		addr:       addr.String(),
 		datacenter: dc,
 		rack:       rack,
-		tokens:     tokens,
 	}, nil
 }
 
+func cqlRowIntoStrategy(r frame.Row) (*strategy, error) {
```
mmatczuk, line 326: Also applies to cqlRowIntoNode.
It's more idiomatic to say what you create from what i.e. readStrategyFromRow.

transport/cluster.go
```diff
@@ -282,6 +489,11 @@ func (c *Cluster) tryReopenControl() {
 		c.control.Close()
 		c.control = control
 	}
+	// Discards request that came to us while reopening.
+	select {
+	case <-c.reopenControlChan:
+	default:
+	}
```
mmatczuk, line 496: Extract to function.

transport/hostselectionpolicy.go
```diff
@@ -0,0 +1,158 @@
+package transport
```
mmatczuk, line 1: I suggest to rename this file to policy

transport/cluster.go
```diff
@@ -100,9 +172,18 @@ func (c *Cluster) refreshTopology() error {
 	}
 	// If node is present in both maps we can reuse its connection pool.
 	old := c.Peers()
-	m := PeerMap{}
-	for _, v := range rows {
-		n, err := cqlRowIntoNode(v)
+	peers := make(PeerMap)
+	dcRacks := make(DCRacksMap)
+	seenRacks := make(map[string]bool) // Key is rack name appended to its dc name.
```
Kulezi, line 177: Is this safe? It could confuse two racks as the same one, for example:
DC = "DC1", Rack = "11"
DC = "DC11", Rack = "1"
would both produce "DC111".
It's unlikely but maybe we should avoid it.

transport/hostselectionpolicy.go
```diff
@@ -0,0 +1,158 @@
...
+		}
+	}
+}
+
+type tokenAwarePolicy struct {
+	wrapperPolicy WrapperPolicy
+}
+
+func newTokenAwarePolicy(pw WrapperPolicy) tokenAwarePolicy {
+	return tokenAwarePolicy{wrapperPolicy: pw}
```
Kulezi, line 89: pw -> wp?

transport/cluster.go
Michal-Leszczynski: Token aware strategies make decisions about taking specific node based on already taken nodes.
Since Simple strategy only cares about choosing unique nodes (which perhaps is not even necessary, because we believe in Scylla and the chances of getting incomplete information about topology is rather slim), maybe we should give up passing **wanted** function and just hard code it inside **ringRange** (name will be changed)?

transport/cluster.go
Michal-Leszczynski: It's not really necessary.
On the other hand, it seems logical to discard refresh topology requests that were sent during refreshing, because they are outdated.

transport/cluster.go
Michal-Leszczynski: I added "?" separator between DC and rack name, so now I don't think that this situation is likely enough to take it into consideration. Do you think it would still be better to just iterate through all nodes and check manually if a given combination of DC and rack has already been encountered?

transport/cluster.go
```diff
@@ -3,39 +3,128 @@ package transport
...
+
+// JCN stands for Java Class Name.
+const (
+	networkTopologyJCN strategyClass = "org.apache.cassandra.locator.NetworkTopologyStrategy"
+	simpleJCN          strategyClass = "org.apache.cassandra.locator.SimpleStrategy"
+	localJCN           strategyClass = "org.apache.cassandra.locator.LocalStrategy"
+	networkTopology    strategyClass = "NetworkTopologyStrategy"
+	simple             strategyClass = "SimpleStrategy"
+	local              strategyClass = "LocalStrategy"
+)
```
mmatczuk, line 63: Can we use more idiomatic names like `networkTopologyStrategy`?

transport/cluster.go
```diff
@@ -118,18 +212,43 @@ func (c *Cluster) refreshTopology() error {
...
 		}
 	}
-	c.setPeers(m)
+
+	c.setTopology(t)
+	drainChan(c.refreshChan)
 	return nil
 }
 
+func allocTopology() *topology {
```
mmatczuk, line 243: newTopology?

mmatczuk (commented):
Looks good some last minute minor comments.

PR #177, opened 2022-03-27
Transport: Cluster, HostSelectionPolicy: Added keyspace

Neofine:


transport/cluster.go
```diff
@@ -238,6 +306,59 @@ func cqlRowIntoNode(r frame.Row) (*Node, error) {
...
+		return &strategy{
+			strat: networkTopology,
+			dcRF:  dcRFs,
+		}, nil
+	case s == "org.apache.cassandra.locator.LocalStrategy" || s == "LocalStrategy":
+		return &strategy{
+			strat: local,
+			rf:    1,
+		}, nil
+	default:
```
Michal-Leszczynski, line 353: Since we aren't supporting custom strategies, maybe we should return error, when unknown strategy appears (just for now)?

transport/cluster.go
```diff
@@ -36,6 +40,34 @@ type topology struct {
...
+
+const (
+	simple          stratType = 0
+	networkTopology           = 1
+	local                     = 2
+	other                     = 3
+)
+
+type strategy struct {
+	strat stratType
```
Michal-Leszczynski, line 63: Since linter does not like shorter version, maybe we should use whole strategy?

transport/cluster.go
```diff
@@ -238,6 +306,59 @@ func cqlRowIntoNode(r frame.Row) (*Node, error) {
...
+			}
+			dcRFs[k] = repFactor(rf)
+		}
+		return &strategy{
+			strat: networkTopology,
+			dcRF:  dcRFs,
+		}, nil
+	case s == "org.apache.cassandra.locator.LocalStrategy" || s == "LocalStrategy":
+		return &strategy{
+			strat: local,
```
Michal-Leszczynski, line 350: Maybe we should set strat to simple right away?

transport/cluster.go
Neofine: Screenshot from 2022-03-29 18-29-23

That's keyspace query returned from my scylla when 'make scylla-up' is called, the last row already contains unknown by us strategy, we could ignore it for now? But having additional information won't hurt us either.

transport/cluster.go
Neofine: Maybe just 'type'? Looking at current strategy type accessing it would be more readable:
Now accessing if rom host selecttion policy looks like:
```go 
if qi.topology.keyspaces[qi.ksName].strategy.strat == ...
```
Your proposition
```go 
if qi.topology.keyspaces[qi.ksName].strategy.strategy == ...
```
With 'type':
```go 
if qi.topology.keyspaces[qi.ksName].strategy.type == ...
```
Having already such a long access maybe making a function out of it would be nice, what do you think @Michal-Leszczynski ?

Neofine: Maybe let's do it like:
```go
type strategy struct {
	stratType
        ...
```
accessing strategy type from host selection policy looks now like:
```go
if qi.topology.keyspaces[qi.ksName].strategy.strat == simple {
	...
```
after this change would be:
```go
if qi.topology.keyspaces[qi.ksName].strategy.stratType == simple {
	...
```
Having such a long access line how would you think if we add a function to do that @Michal-Leszczynski ?

Neofine: I have applied your suggestion, thanks!

Michal-Leszczynski (approved):

Issue #176, opened 2022-03-27
frame: CqlValue - add map[string]string type


Neofine:
Needed for #152 to let us query keyspace values.

PR #178, opened 2022-03-28
Hg/cql as map

Choraden:
Wanted to see how bad generic solution to parse cql as map would be in comparison to strongly typed functions.
Here are the results:
```
go test ./cqlasmapbenchmark_test.go -bench=. -benchmem -benchtime=5s
goos: darwin
goarch: amd64
cpu: Intel(R) Core(TM) i7-7700HQ CPU @ 2.80GHz
BenchmarkGenericMapInt32Int32-8   	  103459	     60339 ns/op	   36097 B/op	    1498 allocs/op
BenchmarkMapInt32Int32-8          	  142498	     41311 ns/op	   30137 B/op	       8 allocs/op
PASS
```
Please let me know if you guys see some easy way to optimise it.

### DIFF
--- a/experiments/cqlasmapbenchmark_test.go
+++ b/experiments/cqlasmapbenchmark_test.go
@@ -1,0 +1,134 @@
+package experiments
+
+import (
+	"fmt"
+	"testing"
+
+	"scylla-go-driver/frame"
+)
+
+func CqlAsMap[K comparable, V any](c frame.CqlValue) (map[K]V, error) {
+	if c.Type.ID != frame.MapID {
+		return nil, fmt.Errorf("is not of map type")
+	}
+
+	keyType := c.Type.Map.Value.ID
+	valueType := c.Type.Map.Key.ID
+
+	var buf frame.Buffer
+	buf.Write(c.Value)
+
+	size := buf.ReadInt()
+
+	m := make(map[K]V, size)
+
+	var key interface{}
+	var value interface{}
+
+	for i := frame.Int(0); i < size; i++ {
+		switch keyType {
+		case frame.IntID:
+			key = buf.ReadInt()
+		default:
+			return nil, fmt.Errorf("unsupported key type")
+		}
+
+		switch valueType {
+		case frame.IntID:
+			value = buf.ReadInt()
+		default:
+			return nil, fmt.Errorf("unsupported value type")
+		}
+
+		if validKey, ok := key.(K); ok {
+			if validValue, ok2 := value.(V); ok2 {
+				m[validKey] = validValue
+				continue
+			}
+		}
+
+		return nil, fmt.Errorf("data can not be parsed to a specified type")
+	}
+
+	return m, buf.Error()
+}
+
+func CqlAsMapInt32Int32(c frame.CqlValue) (map[int32]int32, error) {
+	if c.Type.ID != frame.MapID {
+		return nil, fmt.Errorf("is not of map type")
+	}
+
+	if c.Type.Map.Key.ID != frame.IntID {
+		return nil, fmt.Errorf("invalid key type")
+	}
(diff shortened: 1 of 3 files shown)


Choraden:
@Kulezi I've come up with some idea. It might be helpful with #176 please take a look.

experiments/cqlasmapbenchmark_test.go
```diff
@@ -0,0 +1,134 @@
...
+	}
+
+	return frame.CqlValue{
+		Type: &frame.Option{
+			ID:  frame.MapID,
+			Map: &frame.MapOption{Key: frame.Option{ID: frame.IntID}, Value: frame.Option{ID: frame.IntID}},
+		},
+		Value: buf.Bytes(),
+	}
+}
```
Kulezi, line 100: Protocol: A [int] n indicating the number of key/value pairs in the map, followed by
  n entries.  Each entry is composed of two [bytes] representing the key
  and value.

So here you would actually need to do:
```go
buf.WriteInt(4)
buf.WriteInt(i)
buf.WriteInt(4)
buf.WriteInt(i)
````

**Comment by Kulezi** [comment]
After correcting your functions to use correct map serialization i compared them to my implementation to also see, if using buffer doesn't impact performance in this simple case.
<details>
<summary> Code </summary>

```go
func (c CqlValue) AsInt32Map() (map[int32]int32, error) {
	if c.Type.ID != MapID {
		return nil, fmt.Errorf("%v is not a map", c)
	}
	if c.Type.Map.Key.ID != IntID && c.Type.Map.Value.ID != IntID {
		return nil, fmt.Errorf("%v is not a map[int32]int32", c)
	}

	raw := c.Value
	if len(raw) < 4 {
		return nil, fmt.Errorf("not enough bytes")
	}

	n := int(binary.BigEndian.Uint32(raw))
	raw = raw[4:]
	if len(raw) != 16*n {
		return nil, fmt.Errorf("not enough bytes")
	}

	res := make(map[int32]int32, n)
	for i := 0; i < n; i++ {
		key := int32(binary.BigEndian.Uint32(raw[4:8]))
		value := int32(binary.BigEndian.Uint32(raw[12:16]))
		res[key] = value

		raw = raw[16:]
	}
	return res, nil
}


func BenchmarkMapInt32Int32PP(b *testing.B) {
	c := mapint32int32AsCql(n)

	b.ResetTimer()
	for i := 0; i < b.N; i++ {
		m, err := c.AsInt32Map()
		if err != nil {
			b.Fatal(err)
		}

		if int32(len(m)) != n {
			b.Fatalf("invalid length %v", len(m))
		}
	}
}

```
</details>

Results on my machine:
```
goos: linux
goarch: amd64
cpu: Intel(R) Core(TM) i7-6700HQ CPU @ 2.60GHz
BenchmarkGenericMapInt32Int32-8            65420             93640 ns/op           44289 B/op       1498 allocs/op
BenchmarkMapInt32Int32-8                   85294             70556 ns/op           38329 B/op          8 allocs/op
BenchmarkMapInt32Int32PP-8                183400             37013 ns/op           21945 B/op          7 allocs/op
PASS
```

I think that without getting rid of interface{} we won't see better performance and I don't see a workaround that can do this.
So i guess I'll stick with my current approach for #176, and leave thinking about a general solution for when we will need it.

mmatczuk:
No interface{} please.

Choraden:
It was just an experiment, as Paweł said let's stick with naive approach for now. 
Generic code had shown that the ghaction and linter settings has to be adjusted, which I have done in this pr. Do we want to merge these commits? Eventually I can create separate pr with linter changes. 
@mmatczuk

mmatczuk:
Thanks! I was just looking to fix the same thing today!

PR #180, opened 2022-03-29
transport: conn test, fixed massive query test

Choraden:
- TestConnMassiveQueryIntegration didn't work for large n. Changed test to firstly insert row of given id and then query it.
- N is set to maxStreamID.


mmatczuk:
The same issues apply to the benchmark, i.e. the queries time out due to semaphore on Scylla end.

PR #181, opened 2022-03-30
Transport: Compression added

### LINKED ISSUE #106: transport: add support for compression

Neofine:
- ConnReader can now decompress frame body on recv()
- Added support for snappy and lz4

Fixes #106

### DIFF
--- a/frame/buffer.go
+++ b/frame/buffer.go
@@ -20,6 +20,10 @@ type Buffer struct {
 	readErr error
 }
 
+func (b *Buffer) BytesBuffer() *bytes.Buffer {
+	return &b.buf
+}
+
 func (b *Buffer) Bytes() []byte {
 	return b.buf.Bytes()
 }
--- a/transport/compress.go
+++ b/transport/compress.go
@@ -1,0 +1,148 @@
+package transport
+
+import (
+	"bytes"
+	"encoding/binary"
+	"errors"
+	"io"
+
+	"github.com/klauspost/compress/s2"
+	"github.com/mmatczuk/scylla-go-driver/frame"
+	"github.com/pierrec/lz4/v4"
+)
+
+var (
+	ErrComprCorrupt     = errors.New("compression: corrupt input")
+	ErrComprUnknown     = errors.New("compression: unknown algorithm")
+	errComprUnspecified = errors.New("compression: unspecified algorithm")
+)
+
+type compr struct {
+	buf    bytes.Buffer // Holds the data to compress/decompress from source
+	tmp    bytes.Buffer // Data should be (de)compressed to tmp before writing it to its destination.
+	header []byte
+
+	// codec compresses/decompresses read bytes from buf to tmp.
+	codec func(read int) (written int, err error)
+	// Used in lz4 encoding.
+	c lz4.Compressor
+}
+
+func (c *compr) compress(dst io.Writer, src *bytes.Buffer) (written int64, err error) {
+	c.buf = *src
+	buf := c.buf.Bytes()
+	read := len(c.buf.Bytes())
+
+	if read <= 9 {
+		nr, err := dst.Write(buf[:read])
+		return int64(nr), err
+	}
+	_ = copy(c.header, buf[:9])
+	c.header[1] |= frame.Compress
+
+	var n int
+	if n, err = c.codec(read); err != nil {
+		return 0, err
+	}
+
+	nrh, erh := dst.Write(c.header)
+	if erh != nil {
+		return int64(nrh), erh
+	}
+
+	nr, err := dst.Write(c.tmp.Bytes()[:n])
+	return int64(nr + nrh), err
+}
+
+func (c *compr) decompress(dst io.Writer, src *io.LimitedReader) (written int64, err error) {
+	c.buf.Reset()
+	read := int(src.N)
+
+	_, err = io.Copy(&c.buf, src)
+	if err != nil {
+		return 0, err
+	}
+
+	var n int
+	if n, err = c.codec(read); err != nil {
+		return 0, err
+	}
+
+	nr, err := dst.Write(c.tmp.Bytes()[:n])
+	return int64(nr), err
+}
+
+// newCompr returns a new compressor with respect to whether it should compress or decompress data.
+func newCompr(compress bool, algo frame.Compression, bufSize int) (*compr, error) {
+	c := new(compr)
+	c.buf.Grow(bufSize)
+	switch algo {
+	case frame.Snappy:
+		if compress {
+			c.codec = c.encodeSnappy
+			c.header = make([]byte, 9)
+			c.tmp.Grow(bufSize * bufSize / s2.MaxEncodedLen(bufSize))
+		} else {
+			c.codec = c.decodeSnappy
+			c.tmp.Grow(s2.MaxEncodedLen(bufSize))
+		}
+	case frame.Lz4:
+		if compress {
+			c.codec = c.encodeLz4
+			c.header = make([]byte, 9)
+			c.tmp.Grow(bufSize * bufSize / lz4.CompressBlockBound(bufSize))
+		} else {
+			c.codec = c.decodeLz4
+			c.tmp.Grow(lz4.CompressBlockBound(bufSize))
+		}
+	default:
+		return nil, ErrComprUnknown
+	}
+	return c, nil
+}
+
+func (c *compr) decodeSnappy(read int) (int, error) {
+	buf := c.buf.Bytes()
+	n, err := s2.DecodedLen(buf)
+	if err != nil {
+		return 0, err
+	}
+	c.tmp.Grow(n)
+	if _, err := s2.Decode(c.tmp.Bytes(), buf[:read]); err != nil {
+		return 0, err
+	}
+	return n, nil
+}
+
+func (c *compr) encodeSnappy(read int) (int, error) {
+	c.tmp.Grow(s2.MaxEncodedLen(read - 9))
+	enc := s2.EncodeSnappy(c.tmp.Bytes(), c.buf.Bytes()[9:read])
+	binary.BigEndian.PutUint32(c.header[5:9], uint32(len(enc)))
+	return len(enc), nil
+}
+
+func (c *compr) decodeLz4(read int) (int, error) {
+	if read < 4 {
+		return 0, ErrComprCorrupt
+	}
+
+	buf := c.buf.Bytes()
+	n := int(binary.BigEndian.Uint32(buf[0:4]))
+	c.tmp.Grow(n)
+	if _, err := lz4.UncompressBlock(buf[4:read], c.tmp.Bytes()[:n]); err != nil {
+		return 0, err
+	}
+	return n, nil
+}
+
+func (c *compr) encodeLz4(read int) (int, error) {
+	n := lz4.CompressBlockBound(read - 9)
+	c.tmp.Grow(n + 4)
+	n, err := c.c.CompressBlock(c.buf.Bytes()[9:read], c.tmp.Bytes()[4:(n+4)])
+	if n == 0 || err != nil {
+		return 0, ErrComprCorrupt
+	}
+	binary.BigEndian.PutUint32(c.tmp.Bytes()[:4], uint32(read)-9)
+	binary.BigEndian.PutUint32(c.header[5:9], uint32(n)+4)
+	return n + 4, nil
+}
--- a/transport/compress_test.go
+++ b/transport/compress_test.go
@@ -1,0 +1,158 @@
+package transport
+
+import (
+	"bytes"
+	"encoding/binary"
+	"io"
+	"math/rand"
+	"testing"
+
+	"github.com/klauspost/compress/s2"
+	"github.com/mmatczuk/scylla-go-driver/frame"
+	"github.com/pierrec/lz4/v4"
+)
+
+type comprData struct {
+	name             string
+	uncompressed     []byte
+	compressedSnappy []byte
+	compressedLz4    []byte
+}
+
+func getRandomBytes(n int) []byte {
+	out := make([]byte, n)
+	for i := 9; i < n; i++ {
+		out[i] = byte(rand.Intn(255))
+	}
+	return out
+}
+
+func TestCompression(t *testing.T) {
+	t.Parallel()
+	testCases := []comprData{
+		{
+			name:         "utf-8",
+			uncompressed: append(make([]byte, 9), []byte("Hello World! こんにちは世界! ¡Hola, Mundo!")...),
+		},
+		{
+			name:         "64kb",
+			uncompressed: getRandomBytes(1<<16 + 9),
+		},
+		{
+			name:         "1mb",
+			uncompressed: getRandomBytes(1<<20 + 9),
+		},
+		{
+			name:         "8mb",
+			uncompressed: getRandomBytes(1<<23 + 9),
+		},
+		{
+			name:         "64mb",
+			uncompressed: getRandomBytes(1<<26 + 9),
+		},
+		{
+			name:         "256mb",
+			uncompressed: getRandomBytes(1<<28 + 9),
+		},
+	}
+	for i, c := range testCases {
+		testCases[i].compressedSnappy = s2.Encode(nil, c.uncompressed[9:])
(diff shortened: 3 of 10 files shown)


transport/conn.go
```diff
@@ -327,12 +337,12 @@ func OpenConn(addr string, localAddr *net.TCPAddr, cfg ConnConfig) (*Conn, error
...
 	}
 
-	return WrapConn(tcpConn)
+	return WrapConn(tcpConn, cfg.Compression)
 }
 
 // WrapConn transforms tcp connection to a working Scylla connection.
 // If error and connection are returned the connection is not valid and must be closed by the caller.
-func WrapConn(conn net.Conn) (*Conn, error) {
+func WrapConn(conn net.Conn, cmpr frame.Compression) (*Conn, error) {
```
mmatczuk, line 345: Let's pass the whole config here.

mmatczuk:
What about requests compression?

mmatczuk:
The implementation should use the streaming APIs i.e. io.Reader and io.Writer

Instead of that
```
	// Read body
	if _, err := io.CopyN(c.bufw, c.conn, int64(r.Header.Length)); err != nil {
		r.Err = fmt.Errorf("read body: %w", err)
		return r
	}
	if r.Header.Flags&frame.Compress != 0 {
		if err := c.decompress(&c.buf); err != nil {
			r.Err = fmt.Errorf("decompress body: %w", err)
			return r
		}
	}
```

The `io.CopyN` dst should be the decorating Reader from s2.

Important note we want to avoid allocating those readers for every request and they must be reused.

Same logic applies to writes.

transport/conn.go
```diff
@@ -488,3 +509,25 @@ func (c *Conn) Close() {
...
+	sz := uint32(buf.ReadInt())
+	from := buf.Bytes()
+	dst := make([]byte, sz)
+	_, err := lz4.UncompressBlock(from, dst)
+	if err != nil {
+		return err
+	}
+	buf.Save(dst)
+	return nil
+}
```
mmatczuk, line 533: You can move any compress related funcs to compress.go

transport/conn_integration_test.go
```diff
@@ -160,3 +160,89 @@ func TestCloseHangingIntegration(t *testing.T) {
...
+		t.Fatal(err)
+	}
+	return &connTestHelper{t: t, conn: conn}
+}
+
+func TestSnappyCompressionIntegration(t *testing.T) {
+	h := newConnCompressionTestHelper(t, frame.Snappy)
+	h.applyFixture()
+
+	defer h.close()
```
mmatczuk, line 180: It would make more sense to refactor all the test code to a helper function and call that function in a test (named subtest maybe) with compression as a parameter.

mmatczuk (dismissed):
See the comments inline.

Neofine:
Coding progress is little on this branch because it took me quite a long time to figure out that neither s2.Reader nor snappy.Reader works here.

Both of these readers expect to see chunk stream identifier firstly after they start working that means the exact series of bytes is expected when starting communication:

```
 0xff 0x06 0x00 0x00 0x73 0x4e 0x61 0x50 0x70 0x59
```
Which scylla does not send for obvious reasons.

The problems don't end there, Readers seem to expect first byte as chunk identifier (which tells what data to expect from chunk) and 3 consecutive more to be length of our data.

From my testing first byte in body is the length of data chunk, so our specifications don't match (I haven't checked bodies of greater length than single byte value).

The only solution to this problem I see is to make our own Reader (I haven't gone in-depth with Writers yet), which
uses snappy.Decode in its body, I think this wouldn't be far from original s2.Reader and snappy.Reader but
it would have less features and constraints.

What do you think?

mmatczuk:
Dear @Neofine please note the `ReaderOption` that you add to new reader.
Namely `ReaderIgnoreStreamIdentifier`, we also want `ReaderAllocBlock` to limit the decompression buffer size.
Let me know if that helps.

Neofine:
Compression couldn't be done with neither s2 nor lz4 Readers/Writers because of their funky signature
i.e. s2.Reader needs also crc32 hash of uncompressed data so we can't make our own translation writer.

In compress.go I have added writers and readers to snappy and lz4. 
I am not very satisfied with the result as readers can't be chained with other readers, now we have do:

```go
for {
	if _, err := io.CopyN(c.bufw, c.dec, int64(r.Header.Length)); err != nil {
		if err.Error() != "END" {
			r.Err = fmt.Errorf("read body: %w", err)
			return r
		} else {
			break
		}
	}
}
```
to copy uncompressed data into c.bufw. And because of reader restrictions with my knowledge it can't be done much prettier (we can return n = -1 instead of err = "END" but for loop needs to be there).

Writers seem easy in nature and chainable.

Both snappy and lz4 work on simple test cases. To be sure of its correctness we could add some bigger tests.

Neofine:
Compression couldn't be done with neither s2 nor lz4 Readers/Writers because of their funky signature
i.e. s2.Reader needs also crc32 hash of uncompressed data so we can't make our own translation writer.

In compress.go I have added writers and readers to snappy and lz4. 
Readers not to loop need to return error, so I have added dummy ErrEnd and when performing io.Copy we get this:

```go
if _, err := io.Copy(c.bufw, c.dec); err != nil {
	if !errors.Is(err, ErrEnd) {
		r.Err = fmt.Errorf("read body: %w", err)
		return r
	}
}
```
Any suggestions to make it prettier are appreciated.

Writers seem easy in nature and chainable.

Both snappy and lz4 work on simple test cases. To be sure of its correctness we could add some bigger tests.

transport/compress.go
```diff
@@ -0,0 +1,303 @@
...
+		if err != nil {
+			return 0, r.err
+		}
+
+		if r.err = readFull(r.r, r.buf[:r.l]); r.err != nil {
+			return 0, r.err
+		}
+
+		// TODO: Why we need to do this?
+		if zeroes(r.buf[:r.l]) {
```
Neofine, line 86: This is a band-aid solution.
Scylla when sending `Ready` response seem to also send body of 5 zeroes.
I am not sure of its cause.

frame/buffer.go
```diff
@@ -6,8 +6,11 @@ import (
 )
 
 // CopyBuffer writes data to w until the buffer is drained or an error occurs.
-func CopyBuffer(b *Buffer, w io.Writer) (n int64, err error) {
-	return b.buf.WriteTo(w)
+func CopyBuffer(b *Buffer, w io.Writer) (n int, err error) {
```
mmatczuk, line 9: Why is this changed?

mmatczuk, line 9: Why is this changed?

transport/compress.go
```diff
@@ -0,0 +1,232 @@
...
+	"github.com/klauspost/compress/s2"
+	"github.com/pierrec/lz4/v4"
+)
+
+const maxBlockSize = 4 << 20
+
+var (
+	ErrComprCorrupt     = errors.New("compression: corrupt input")
+	ErrComprUnsupported = errors.New("compression: unsupported input")
+	errComprEnd         = errors.New("end")
```
mmatczuk, line 19: Can we use nil instead of it?

mmatczuk:
Some nice improvements @Neofine.

I pushed a refactoring changes - I wanted to merge this but there is a problem.

The issue is that we have here a growing buffer. I'd like to suggest that following:

- Forget the "babushka" style wrapping of readers and writers
- Add a helper type that would expose `Copy(dst Writer, src Reader) (written int64, err error)`
- The constructor parameters are compress or decompress, type of compression, compression buffer size
- The size of buffer should be in ConnConfig, I suggest default of 64Kb but please double check that
- There must be `buf` and `tmp` `[]byte` fields of fixed size initiated on construction
- The copy would follow [Copy with buffer code from io pkg](https://github.com/golang/go/blob/78fb1d03d39e8357e4790a9f0788ef0a8e7d8ae1/src/io/io.go#L404)
    - Fill the buffer (buf)
    - Call a function: compress or decompress (best done with function pointer)
    - Write tmp
- On the perf side you can detect if reader is `bytes.Buffer` and access the bytes directly no need to copy them again

transport/compress.go
```diff
@@ -0,0 +1,232 @@
...
+	n, err := w.c.CompressBlock(p[9:], w.encoded)
+	if n == 0 || err != nil {
+		return nil, ErrComprUnsupported
+	}
+
+	l := make([]byte, 4, 4)
+	binary.BigEndian.PutUint32(l, uint32(len(p)))
+
+	enc = p[:9]
+	enc = append(enc, l...)
```
mmatczuk, line 228: Would be better to append shifted bytes directly, we do that in other places already.

transport/compress.go
```diff
@@ -0,0 +1,172 @@
...
+	header []byte
+
+	compressor func(int) (int, error)
+	c          lz4.Compressor
+}
+
+func newCompression(compress, snappy bool, bufSize int) *compr {
+	c := &compr{
+		buf: make([]byte, bufSize),
+		tmp: make([]byte, bufSize),
```
mmatczuk, line 32: This should probably be larger / smaller depending on compression type.
Before you calculated the size the temporary buffer.

transport/compress.go
```diff
@@ -0,0 +1,172 @@
...
+	}
+	switch compress {
```
mmatczuk, line 34: Do not use a switch over bool.

transport/compress.go
```diff
@@ -0,0 +1,172 @@
...
+	}
+
+	binary.BigEndian.PutUint32(c.tmp[:4], uint32(read))
+	binary.BigEndian.PutUint32(c.header[5:9], uint32(n)+4)
+	return n + 4, nil
+}
+
+func (c *compr) Copy(dst io.Writer, src io.Reader) (written int64, err error) {
+	var (
+		toRead int
```
mmatczuk, line 106: how about size?

transport/compress.go
```diff
@@ -0,0 +1,172 @@
...
+	ErrComprCorrupt     = errors.New("compression: corrupt input")
+	ErrComprUnsupported = errors.New("compression: unsupported input")
+)
+
+type compr struct {
+	buf    []byte
+	tmp    []byte
+	header []byte
+
+	compressor func(int) (int, error)
```
mmatczuk, line 25: I'd call that process or something since it can compress or decompress.

transport/compress.go
```diff
@@ -0,0 +1,172 @@
...
+				}
+				break
+			}
+			if read == toRead {
+				break
+			}
+		}
+	}
+
+	if len(c.header) != 0 {
```
mmatczuk, line 132: It would be easier to read if c.header != nil

transport/compress.go
```diff
@@ -0,0 +1,172 @@
...
+	}
+
+	var (
+		nrh int
+		erh error
+	)
+	if len(c.header) != 0 {
+		nrh, erh = dst.Write(c.header)
+		if erh != nil {
+			if ok {
```
mmatczuk, line 160: What's that?

Neofine:
I may be mistaken but from my observations scylla doesn't seem to divide data into the actual chunks, even when asked to do so (with `'chunk_length_kb': 64`), after making (not compressed) insertion of 600k characters - approx 0.6 MB I query compressed data and first bytes of chunks from both lz4 and snappy are telling that length of uncompressed data of nearest chunk is 600k.

In both gocql and rust drivers we have respectively:
```go
func (s SnappyCompressor) Decode(data []byte) ([]byte, error) {
	return snappy.Decode(nil, data)
}
```
```rust
...
Compression::Snappy => snap::raw::Decoder::new()
            .decompress_vec(comp_body)
            .map_err(|_| FrameError::FrameDecompression),
...
```

where both functions work on full body of a frame. In golang `snappy.Decode(nil, data)` where `data` consists of two concatenated chunks results in error so gocql surely expects a single big chunk.

s2.Reader has option of reading data chunk by chunk:

```go
chunkLen := int(r.buf[1]) | int(r.buf[2])<<8 | int(r.buf[3])<<16
```

but we resigned from using it as scylla doesn't give us data compatible with it.

With my current understanding of the situation we can't avoid sudden allocations of slices of (in this case) 600k items,  what should we do now?

transport/conn.go
```diff
@@ -113,18 +114,25 @@ func (c *connWriter) send(r request) error {
 	binary.BigEndian.PutUint32(b[5:9], l)
 
 	// Send
-	if _, err := frame.CopyBuffer(&c.buf, c.conn); err != nil {
-		return err
+	var err error
+	if r.Compress {
```
Kulezi, line 118: c.compr being nil probably should return some meaningful error instead of nil

transport/conn.go
```diff
@@ -432,18 +460,27 @@ func validateKeyspace(keyspace string) error {
 	return nil
 }
 
-var startupOptions = frame.StartupOptions{"CQL_VERSION": "3.0.0"}
+const cqlVersion = "3.0.0"
 
-func (c *Conn) init() error {
+func (c *Conn) init(cfg ConnConfig) error {
```
Kulezi, line 465: Why give cfg to a *Conn method when Conn owns it already?

transport/compress.go
```diff
@@ -0,0 +1,148 @@
...
+	enc := s2.Encode(c.tmp.Bytes(), c.buf.Bytes()[9:read])
+	binary.BigEndian.PutUint32(c.header[5:9], uint32(len(enc)))
+	return len(enc), nil
+}
+
+func (c *compr) decodeLz4(read int) (int, error) {
+	buf := c.buf.Bytes()
+	if read < 4 {
+		return 0, ErrComprCorrupt
+	}
```
Kulezi, line 81: check should be done before declaring buf

transport/compress.go
```diff
@@ -0,0 +1,148 @@
...
+	if n == 0 || err != nil {
+		return 0, ErrComprCorrupt
+	}
+	binary.BigEndian.PutUint32(c.tmp.Bytes()[:4], uint32(read)-9)
+	binary.BigEndian.PutUint32(c.header[5:9], uint32(n)+4)
+	return n + 4, nil
+}
+
+func (c *compr) Compress(dst io.Writer, src *bytes.Buffer) (written int64, err error) {
+	c.buf = *src
```
Kulezi, line 103: should c.buf be overwritten like that?

Kulezi: n/m

transport/compress_test.go
```diff
@@ -0,0 +1,158 @@
...
+
+func getRandomBytes(n int) []byte {
+	out := make([]byte, n)
+	for i := 9; i < n; i++ {
+		out[i] = byte(rand.Intn(255))
+	}
+	return out
+}
+
+func TestCompression(t *testing.T) {
```
Michal-Leszczynski, line 30: It seems like random tests / fuzzing could be an interesting addition for testing compression.

transport/compress.go
```diff
@@ -0,0 +1,146 @@
...
+	if n, err = c.codec(read); err != nil {
+		return 0, err
+	}
+
+	nrh, erh := dst.Write(c.header)
+	if erh != nil {
+		return int64(nrh), erh
+	}
+
+	nr, er := dst.Write(c.tmp.Bytes()[:n])
```
Choraden, line 52: How about we wrap this line with some WriteCompressed to hide tmp variable, which is used in codec function. 
What I mean is that codec function silently uses tmp, but it is not documented. I would opt for hiding it completely.

transport/compress.go
```diff
@@ -0,0 +1,146 @@
...
+	if err != nil {
+		return 0, err
+	}
+
+	var n int
+	if n, err = c.codec(read); err != nil {
+		return 0, err
+	}
+
+	nr, er := dst.Write(c.tmp.Bytes()[:n])
```
Choraden, line 70: Same here.

Choraden (commented):
left some comments

transport/compress.go
Kulezi: Added documentation about tmp usage, I believe now it's ok

transport/compress.go
```diff
@@ -0,0 +1,146 @@
...
+	return int64(nr), er
+}
+
+func newCompression(compress bool, algo frame.Compression, bufSize int) (*compr, error) {
```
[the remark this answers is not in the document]
Kulezi: Changed to newCompr to fit the struct name + added comment about the bool.

PR #182, opened 2022-04-02
Added conn info to logs and errors originated from reader and writer

Choraden:


transport/conn.go
```diff
@@ -132,12 +134,12 @@ func (c *connReader) setHandler(h responseHandler) (frame.StreamID, error) {
 	defer c.mu.Unlock()
 
 	if c.closed {
-		return 0, fmt.Errorf("connection closed")
+		return 0, fmt.Errorf("%s - closed", c.info)
```
mmatczuk, line 137:
```suggestion
		return 0, fmt.Errorf("%s closed", c.info)
```

transport/conn.go
```diff
@@ -360,6 +362,9 @@ func WrapConn(conn net.Conn) (*Conn, error) {
 		return c, err
 	}
 
+	c.w.info = c.String()
+	c.r.info = c.String()
+
```
mmatczuk, line 367: This can be moved to the declaration directly.

mmatczuk: That is under r: connReader{

Choraden: I'm afraid this can't be declared before running init() method which discovers shard. Also RemoteAddr() method in conn::String() causes nil pointer dereference when called in r: connReader{ or w: connWriter, since conn is not set.

mmatczuk: I was thinking about a function pointer not a string.

PR #185, opened 2022-04-05
session: add Query result iterator

Kulezi:


mmatczuk:
Maybe we want to Iter on Query.

mmatczuk:
`type Row frame.Row` let's not export raw Row definition in the top level API.
If you really need it use row but this can be inlined imo.

mmatczuk:
Overall looks good.

mmatczuk (approved):

PR #186, opened 2022-04-07
transport: added support for authentication

### LINKED ISSUE #105: Add support for authentication

Michal-Leszczynski:
In gocql and rust they only supported  **PasswordAuthenticator** and used it no matter what approved authenticator was sent in **Authenticate** response.
Do we also only care for this one?

Fixes #105


mmatczuk:
We can do better and add some validation.

mmatczuk (approved):

mmatczuk:
log.Fatal is no go it would kill the whole app using it.

mmatczuk:
Please restore returning errors.

PR #197, opened 2022-04-11
transport: reusing response handlers

Choraden:
- ResponseHandlersAllocator reuses already allocated handlers or allocates new one. 
- Default allocated handlers size was chosen to be 16000 ~ half of all the streamIDs
- Modified connConfig struct to store PreallocatedHandlers size.
- There is no need to drain the handlers. Once the error occurs connection will be closed and channel size is 2, so this will not block.

### DIFF
--- a/session.go
+++ b/session.go
@@ -54,16 +54,18 @@ var (
 		"SERIAL      Consistency = 0x0008\n" +
 		"LOCALSERIAL Consistency = 0x0009\n" +
 		"LOCALONE    Consistency = 0x000A")
-	errNoConnection = fmt.Errorf("no working connection")
+	errNoConnection   = fmt.Errorf("no working connection")
+	errNoPreallocated = fmt.Errorf("preallocated handlers size must be greater than 0")
 )
 
 func DefaultSessionConfig(hosts ...string) SessionConfig {
 	return SessionConfig{
 		Hosts: hosts,
 		ConnConfig: transport.ConnConfig{
-			Timeout:            500 * time.Millisecond,
-			TCPNoDelay:         true,
-			DefaultConsistency: LOCALQUORUM,
+			Timeout:              500 * time.Millisecond,
+			TCPNoDelay:           true,
+			DefaultConsistency:   LOCALQUORUM,
+			PreallocatedHandlers: 16000,
 		},
 	}
 }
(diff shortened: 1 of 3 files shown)


Choraden:
It looks like the gain is not that much or It was difficult to me to measure the time in a reliable way on Darwin with docker, so I have a great request to @Neofine if you could run benchmark on your setup, I would be so so pleased :D

session.go
```diff
@@ -54,16 +54,18 @@ var (
...
 	return SessionConfig{
 		Hosts: hosts,
 		ConnConfig: transport.ConnConfig{
-			Timeout:            500 * time.Millisecond,
-			TCPNoDelay:         true,
-			DefaultConsistency: LOCALQUORUM,
+			Timeout:              500 * time.Millisecond,
+			TCPNoDelay:           true,
+			DefaultConsistency:   LOCALQUORUM,
+			PreallocatedHandlers: 16000,
```
mmatczuk, line 68: I do not think this needs to be configured.
There is a task to put all constants to ConnConfig so it would be configured later.
Keeping it the same size as request chan size makes sense.

transport/conn.go
```diff
@@ -121,34 +119,39 @@ type connReader struct {
...
-func (c *connReader) freeHandler(streamID frame.StreamID) {
+func (c *connReader) freeHandler(streamID frame.StreamID, h responseHandler) {
 	c.mu.Lock()
-	c.s.Free(streamID)
+	defer c.mu.Unlock()
+
 	delete(c.h, streamID)
-	c.mu.Unlock()
+	c.s.Free(streamID)
+	c.handlers.yield(h)
```
mmatczuk, line 154: free?

session.go
mmatczuk: Or maybe not, I'd use 1k as a starting point.

mmatczuk:
The cost of allocating all of them is 2,5MiB per connection that's substantial not impossible not very smart either.

```
package main

import (
	"fmt"
	"math"
	"runtime"

	"github.com/mmatczuk/scylla-go-driver/frame"
)

type response struct {
	frame.Header
	frame.Response
	Err error
}

type responseHandler chan response

func main() {
	const v = math.MaxInt16
	vv := [v]responseHandler{}
	for i := 0; i < v; i++ {
		vv[i] = make(responseHandler, 2)
	}

	var ms runtime.MemStats
	runtime.ReadMemStats(&ms)
	fmt.Printf("%dkB %dkB", ms.HeapInuse/1024, ms.Alloc/1024)
}
```

I do not think that keeping a fixed amount makes sense. Consider a properly done pool https://github.com/rclone/rclone/blob/master/lib/pool/pool.go.
- It can grow
- And it can shrink
- In our case we also need the returned items to drained

PR #201, opened 2022-04-12
session: uses host selection policy to pick connections

Choraden:


mmatczuk:
I asss ume you figured out that 	`PkIndexes         []frame.Short` can be committed in clone because they do not change.
If so it's worth noting it.

mmatczuk:
```
const (
	RoundRobin = iota
	TokenAware
	DCAware
)
```

This does not really make any sense, user should be able to set `HostSelectionPolicy` in config.

mmatczuk:
Could we rename `hsp` to `policy` ex. `q.session.hsp.PlanIter(info)` vs. `q.session.policy.PlanIter(info)`

mmatczuk:
```
func (q *Query) info(token transport.Token, tokenAware bool) transport.QueryInfo {
	if tokenAware {
		// TODO: Will the driver support using different keyspaces than default?
		if info, err := q.session.cluster.NewTokenAwareQueryInfo(token, ""); err == nil {
			return info
		}
	}

	return q.session.cluster.NewQueryInfo()
}
```

Should we have some error propagation here?

mmatczuk:
@Choraden I'm merging this please note the comments above

PR #202, opened 2022-04-12
transport: replaced interface btree with generic one

### LINKED ISSUE #187: btree use the generic version

Michal-Leszczynski:
Fixes #187


Michal-Leszczynski:
Is 'replace' directive (go.mod) considered to be a bad practise?

mmatczuk:
It's legit.

PR #205, opened 2022-04-14
frame: added fuzzing for tests

### LINKED ISSUE #191: Add fuzzing to framer / or connection
Read first https://go.dev/doc/fuzz/

Michal-Leszczynski:
Changes:
```
- added fuzzing for buffer tests
- added fuzzing for cql value tests
  - both of those are based on the fact that deserialize(serialize(x)) == x
    and they test behaviour for correct input
- added fuzzing for requests and responses
  - the last checks if writing and reading frames doesn't cause program to crash
    for random data (mostly incorrect input)
- discovered and fixed vulnerability in buffers 'readCopy' method, where
  even for corrupted or enormously large int, driver would still try to allocate
  slice of given length causing program to crash
```
Fixes #191

### DIFF
--- a/frame/buffer_fuzz_test.go
+++ b/frame/buffer_fuzz_test.go
@@ -1,0 +1,202 @@
+package frame
+
+import (
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+)
+
+// Fuzzing tests are based on:
+//   - data            == deserialize(serialize(data))
+//   - serialize(data) == serialize(deserialize(serialize(data)))
+// with the exception of comparing serialized maps, which is nondeterministic because of
+// nondeterministic iteration over build-in map.
+//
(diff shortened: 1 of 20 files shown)


Michal-Leszczynski:
I mostly kept fuzz tests in test files, but for the buffer i created separate file, since it has separate files for testing write and read behavior. Is this filie division acceptable?

Michal-Leszczynski:
I would like to add comments about what is the point of specific fuzzing (like in commit message), but I'm not sure about placement of those comments. Should they be in the docs file, or should the be at the beginning of fuzz tests in each file? If the latter is better, then should those comments be replicated across all test files in request and response package as well?

Michal-Leszczynski:
I believe that testing requests and responses on random / incorrect input is enough, since they are the point of communication with outer world and they use buffer and cqlValue parsing functions (so in a way they are also tested). Am I correct?

PR #209, opened 2022-04-21
session improve

Choraden:
Addressed comments in #201


session.go
```diff
@@ -195,6 +165,22 @@ func (s *Session) Prepare(content string) (Query, error) {
 	}, err
 }
 
+func (s *Session) NewRoundRobinPolicy() transport.HostSelectionPolicy {
```
mmatczuk, line 168: @Choraden I'm merging that but FYI #211

PR #210, opened 2022-04-21
Ml/imporving cluster ip handling

### LINKED ISSUE #175: Support using different than default scylla port

Michal-Leszczynski:
Fixes #175


Michal-Leszczynski:
@Neofine Let me know if this PR helps with connecting to cloud.

transport/cluster.go
```diff
@@ -211,7 +217,8 @@ func (c *Cluster) refreshTopology() error {
 				n.pool = pool
 			}
 		}
-
+		// Every encountered node becomes known host for future use.
+		c.knownHosts[n.addr] = true
```
mmatczuk, line 221: It's better to use struct{} instead of bool.

transport/pool.go
```diff
@@ -129,12 +128,14 @@ type PoolRefiller struct {
 	active int
 }
 
+// Here addr will consist only of IP, because init is called on addresses
+// gathered from system.peers or system.local tables.
 func (r *PoolRefiller) init(addr string) error {
```
mmatczuk, line 133: I suggest you now replace `addr` with `host` in this file and its test.

transport/conn_integration_test.go
```diff
@@ -37,7 +40,10 @@ type connTestHelper struct {
 }
 
 func newConnTestHelper(t testing.TB) *connTestHelper {
-	conn, err := OpenConn(TestHost, nil, ConnConfig{Timeout: 500 * time.Millisecond})
+	conn, err := OpenConn(TestHost, nil, ConnConfig{
```
mmatczuk, line 43: Let's move creation of the default conn config to transport pkg.

mmatczuk:
Overall looks good.

mmatczuk:
Fix the comments and we can merge.

transport/pool.go
Michal-Leszczynski: Can we have the following name convention?
- **host** is always IP only
- **addr** can be IP only or IP with port
We can still open connection to both host and addr, because in case of lacking port, it is automatically set to the default one.

Michal-Leszczynski: On the other hand, this convention won't hold in cluster, where we are asking Scylla for **rpc_address** and getting IP only.
Moreover, **Refiler** has field named **addr**, so should it be renamed to **host** even if it can contain port?

PR #219, opened 2022-04-25
frame, response: added extra checking before allocating slice

### LINKED ISSUE #212: fix reading error found by fuzzing
Data added in 9eb030b7

Michal-Leszczynski:
When allocating a slice for reading purposes, we check if it has valid length and if error hasn't been encountered before.

Fixes #212

### DIFF
--- a/frame/buffer_read.go
+++ b/frame/buffer_read.go
@@ -39,7 +39,7 @@ func (b *Buffer) readInto(p []byte) {
 	}
 }
 
-const maxSliceSize int = 128_000_000
+const maxBytesSize int = 128_000_000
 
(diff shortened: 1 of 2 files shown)


Michal-Leszczynski:
Im not sure about the slice size limits. Do they make sense?

mmatczuk:
I do not necessarily like intermediate error checking 

```
	if b.readErr != nil {
		return nil
	}
```

what is the purpose of it?

Otherwise looks like a nice hardening the fuzzing found!

Michal-Leszczynski:
I removed most intermediate error checking but I'm still worried about the limits.

PR #225, opened 2022-04-28
Ml/retry policy

Michal-Leszczynski:
Added retry policy and extended handling of Scylla errors.
Currently all errors (except for Scylla error frames) are treated as `IoErrors`.
In Rust they also have `ProtocolErrors` (eg. for unknow frame type).
The difference in terms of retry policy is:
```
IoErros        -> RetryNextNode
ProtocolErrors -> DontRetry
```
Do we have to distinguish them? In special scenarios, unknown frame type can be result of connection error, so maybe we can treat them as `IoErrors`?


frame/types.go
```diff
@@ -109,6 +109,82 @@ const (
 // https://github.com/apache/cassandra/blob/trunk/doc/native_protocol_v4.spec#L502
 type ErrorCode = Int
 
+// See CQL Binary Protocol v4, section 9 for more details.
+// https://github.com/apache/cassandra/blob/trunk/doc/native_protocol_v4.spec#L1046-L1200
+const (
```
Neofine, line 114: Is there a need for errors to be located in frame/types.go? We already have them in errors.go

mmatczuk:
It is OK to have to copies, for clarity put them to a separate file.
The files should be called the same and should have a commend documenting why two copies exists.

mmatczuk:
I think you could do

```
type SyntaxError ScyllaError
```

instead of

```
type SyntaxError struct {
	ScyllaError
}
```

mmatczuk:
`func ParseXXXXError(b *frame.Buffer, err ScyllaError) *XXXError` and return a value instead of a pointer for compatibility with other code.

Having separate but void types for every error make no sense, let's use a common code for the default case.

mmatczuk:
I still needs to be integrated with session.

PR #233, opened 2022-05-06
Pre processing policies

Choraden:
In this approach we prepare plan for given policy during cluster::refreshTopology, then Iter() is costless.


Choraden:
Screenshot 2022-05-06 at 09 33 42

transport/policy.go
```diff
@@ -256,8 +210,30 @@ func (p *NetworkTopologyTokenAwarePolicy) Update(t *topology) {
...
-func (p *NetworkTopologyTokenAwarePolicy) Clone() HostSelectionPolicy {
-	return p // TODO: implement valid Clone method.
+func NewNetworkTopologyTokenAwarePolicy(wp WrapperPolicy) *NetworkTopologyTokenAwarePolicy {
+	return &NetworkTopologyTokenAwarePolicy{
+		WrapperPolicy: wp,
+	}
+}
+
+func (p NetworkTopologyTokenAwarePolicy) Iter(qi QueryInfo, idx int) *Node {
+	start, end := 0, len(p.PreparedNodes)
```
Choraden, line 226: Shouldn't it be 
```go
start, end := 0, len(p.PreparedNodes)-1
```
?

Choraden: Then I believe you can delete this
```go
if start == len(p.PreparedNodes) {
 	start = 0
}
```

transport/cluster_integration_test.go
```diff
@@ -74,7 +74,7 @@ func TestClusterIntegration(t *testing.T) {
 	}
 
 	// There should be at least system keyspaces present.
-	if len(c.topology.Load().(*topology).keyspaces) == 0 {
+	if len(c.topology.Load().(*Topology).keyspaces) == 0 {
```
Choraden, line 77: What's the type assertion for?

session.go
```diff
@@ -100,7 +102,19 @@ func (cfg *SessionConfig) Validate() error {
 type Session struct {
 	cfg     SessionConfig
 	cluster *transport.Cluster
-	policy  transport.HostSelectionPolicy
+	hsp     atomic.Value // HostSelectionPolicy
```
Choraden, line 105: Should we keep the naming as it was? This is hsp -> policy.

transport/policy.go
Michal-Leszczynski: Answers:
- no
- no

transport/cluster_integration_test.go
Michal-Leszczynski: It casts the type.

session.go
Michal-Leszczynski: I changed policy to `hsp`, because you cannot have method and filed of the same name.
We won't really be using `hsp` anyways, we will always call `policy()` method to load it from `atomic.Value`.

transport/policy.go
Michal-Leszczynski: Tokens: [1, 3, 5, 7]
Query: 8
Got: 7
Wanted: 1

transport/cluster_integration_test.go
Choraden: We can always use already implemented method: 
```go
func (c *Cluster) Topology() *Topology {
	return c.topology.Load().(*Topology)
}
```

mmatczuk:
Without this patch I'm getting consistently in 10M mixed

```
2022/05/27 12:26:17 Finished
Benchmark time: 11771 ms
```

Kulezi:
Closing in favor of #247

PR #234, opened 2022-05-06
Request coalescing 

Kulezi:


transport/conn.go
```diff
@@ -44,13 +44,58 @@ type ConnMetrics struct {
...
+
+		time.Sleep(waitTime)
+	}
+}
+
+func (cs *coalescingStrategy) update() {
+	if cs.last.IsZero() {
+		cs.last = time.Now()
+		return
+	}
```
mmatczuk, line 76: Can we init that upfront and remove the if?

transport/conn.go
```diff
@@ -44,13 +44,58 @@ type ConnMetrics struct {
 	InQueue  atomic.Uint32
 }
 
+type coalescingStrategy struct {
```
mmatczuk, line 47: conn is quite full, maybe we can move it to coalescing.go ?

mmatczuk:
Looks nice - this would benefit from some comments thought

transport/conn.go
```diff
@@ -44,13 +44,58 @@ type ConnMetrics struct {
...
+	gaps  [16]time.Duration
```
mmatczuk, line 48: 16 could be a const or a constructor param.

transport/conn.go
```diff
@@ -44,13 +44,58 @@ type ConnMetrics struct {
...
+func (cs *coalescingStrategy) update() {
+	if cs.last.IsZero() {
+		cs.last = time.Now()
+		return
+	}
+
+	gap := time.Since(cs.last)
+	cs.last = time.Now()
+	cs.total += gap - cs.gaps[cs.pos]
+	cs.gaps[cs.pos] = gap
```
mmatczuk, line 81: I'm not sure about this code
- it looks like the array is nit needed and you could do with the last value only
- it looks like in L80 we calculate difference between current and last batch

Kulezi:
Results of running benchmarks on my local machine (Intel® Core™ i7-6700HQ CPU @ 2.60GHz × 8, 16GB RAM):

No wait at all: 10893ms on average
nowait

Waiting for double of moving average of last 16 gaps between sends: 9976 ms on average
movingavg

Fixed 1ms wait: 9373 ms on average
fixed

mmatczuk:
Thanks for the nice work but it seems dumb small sleep works best.
I added it in 25253d73.

PR #236, opened 2022-05-06
cmd/benchtab: fix async mode

mmatczuk:
Async mode was broken - test would never end.
The root cause was copying `nextBatchStart int64` to massAsyncQuery and then getting the address of it.

This patch contains a refactoring
* massAsyncQuery is inlined
* massAsyncXXX is renamed to asyncXXX
* batch-size can be set from command line


mmatczuk:
FYI @Choraden there was a bug in the test code

PR #242, opened 2022-05-08
frame: added fuzzing configurations to result

Neofine:


.golangci.yml
```diff
@@ -76,3 +76,7 @@ issues:
       linters:
         - dupl
         - thelper
+    - path: _test_input\.go
```
Neofine, line 79: Should we break lines at 120 in the input test file or set linter to ignore it?

frame/response/result_fuzz_test_input.go
```diff
@@ -0,0 +1,53 @@
+package response
+
+var rowsResultTests = [][]byte{
```
Neofine, line 3: If there is need plenty more rows results can be added.

Neofine:
Is there any other important part of frame that configurations should be added?

PR #256, opened 2022-07-14
transport: added preprocessed policies

Kulezi:
- removed wrapping of policies
- policy interface is now just Node(QueryInfo, n int), which returns the n-th node of the plan
- created policyInfo struct with Preprocess() method, meant to be called on topology refresh
  to allow calculating all possible query plans in advance,
  its passed to the policy inside QueryInfo for further use.

### DIFF
--- a/transport/cluster.go
+++ b/transport/cluster.go
@@ -33,15 +33,17 @@ type Cluster struct {
 	refreshChan       requestChan
 	reopenControlChan requestChan
 	closeChan         requestChan
+
+	queryInfoCounter atomic.Uint64
 }
 
 type topology struct {
-	peers     peerMap
-	dcRacks   dcRacksMap
-	nodes     []*Node
-	ring      Ring
-	trie      trie
-	keyspaces ksMap
+	localDC    string
+	peers      peerMap
+	dcRacks    dcRacksMap
+	nodes      []*Node
+	policyInfo policyInfo
+	keyspaces  ksMap
 }
 
 type keyspace struct {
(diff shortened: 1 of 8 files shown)


mmatczuk:
Looking at the code I think we can squash the commits to one.

transport/cluster.go
```diff
@@ -110,52 +114,13 @@ func (c *Cluster) NewTokenAwareQueryInfo(t Token, ks string) (QueryInfo, error)
...
-		end = 0
-	}
-
-	return replicaIter{
-		ring:   t.ring,
-		offset: end,
-	}
+// TODO overflow and negative modulo.
+func (c *Cluster) generateOffset() int {
+	return int(c.queryInfoCounter.Inc()) - 1
```
mmatczuk, line 119: Why not put it in session?

transport/node.go
```diff
@@ -51,3 +53,44 @@ type Ring []RingEntry
...
+	}
+
+	return ret
+}
+
+// tokenPos returns the position of first node with token larger than given, 0 if there wasn't one.
+func (r Ring) tokenLowerBound(token Token) int {
+	start, end := 0, len(r)
+	for start < end {
+		mid := int(uint(start+end) >> 1)
```
mmatczuk, line 83: nice trick with uint

transport/node.go
```diff
@@ -51,3 +53,44 @@ type Ring []RingEntry
...
+		if r[mid].token < token {
+			start = mid + 1
+		} else {
+			end = mid
+		}
+	}
+
+	if end >= len(r) {
+		end = 0
```
mmatczuk, line 92: Why 0

transport/cluster.go
```diff
@@ -33,15 +33,17 @@ type Cluster struct {
 	refreshChan       requestChan
 	reopenControlChan requestChan
 	closeChan         requestChan
+
+	queryInfoCounter atomic.Int64
```
mmatczuk, line 37: Use uint64 and offset should also be uint64

Issue #250, opened 2022-07-01
No contexts?


martin-sucha:
NewSession, Session.Prepare, Query.Execute do network queries, but the methods do not take context. AsyncExec does not return the query result. How should users handle abandoning a running query once request context deadline is reached? I could call Query.Execute in a separate goroutine, but keeping a whole goroutine open just to wait until the server responds to a stream that the user is no longer interested in seems excessive. For NewSession and Session.Prepare there is no way to cancel the network requests.

PR #263, opened 2022-07-20
TLS support

Kulezi:
This pull request adds basic TLS support to the driver.
- Added *tlsConfig to SessionConfig so if *tls.Config is provided in SessionConfig, all connections will use TLS.
- Updated docker-compose.yml to use a scylla instance supporting encryption
- Added some integration tests to verify if tls works, and if I created valid certificates.

For now i implemented it as minimal as possible, but maybe we should use a HostDialer interface (https://github.com/gocql/gocql/pull/1629, https://github.com/scylladb/gocql/pull/98)?
Right now the driver doesn't support Contexts, I'd add it with a PR resolving #250


Choraden (approved):

session_integration_test.go
```diff
@@ -230,3 +233,112 @@ func TestSessionIterIntegration(t *testing.T) { // nolint:paralleltest // Integr
...
+				InsecureSkipVerify: true,
+			},
+			certPath: certPath,
+			keyPath:  keyPath,
+		},
+		{
+			name: "tls - with client verification",
+			tlsConfig: &tls.Config{
+				RootCAs:            x509.NewCertPool(),
+				InsecureSkipVerify: false,
```
avelanarius, line 268: Out of curiosity: how did you make it pass correctly with client verification - yesterday it wasn't working, what was the problem?

zimnx: Without setting RootCA to DB CA you shouldn't be able to verify certificate, something is fishy :)

testdata/config/scylla.yaml
```diff
@@ -7,4 +7,5 @@ client_encryption_options:
   certificate: /etc/scylla/db.crt
   keyfile: /etc/scylla/db.key
   truststore: /etc/scylla/cadb.pem
-  require_client_auth: true
\ No newline at end of file
+  require_client_auth: true
+
```
piodul, line 11: Nit: this whitespace change should be included in the first commit.

piodul: Same goes for changes in `db.cfg` and `gen_certificates.sh`

session_integration_test.go
Kulezi: > Out of curiosity: how did you make it pass correctly with client verification - yesterday it wasn't working, what was the problem?

The problem was that if InsecureSkipVerify flag is false ServerName must be provided and supplying node's IP address as ServerName wouldn't work with certificates from gocql repo as they don't list any SAN's. I created new certificates with node's IP listed as a SAN, with it it works. I guess it has to do with the fact that default gocql sslOptions default to not doing client verification (setting InsecureSkipVerifyFlag to true).

piodul (approved):

PR #264, opened 2022-07-21
api: use contexts

Kulezi:
This PR changes NewSession, Session.Prepare, Query.Execute, Query.AsyncExecute to take context.Context as an argument.
Context passed to NewSession forces it to close on context cancellation, this is done by:
- checking ctx.Done() in cluster.loop and closing the cluster if ctx is cancelled
- checking if context is cancelled in OpenConn, preventing dialing after context cancellation.
- preventing topology refresh if context is cancelled, and during refresh in calls to openConnPool preventing opening new pools.

NewSession's context also stops requests from being sent to server in connWriter.send after cancellation,
contexts passed to Session.Prepare, Query.Execute, Query.AsyncExecute provide the same behaviour.
Context's cancellation is also checked right before allocating requests StreamID.

Requests that were sent but didn't reach the driver before context cancellation are met with an error response mentioning context cancellation, I'm not sure if this is right or if there should be a bit of additional time to receive the response, as the request may have been sent near context deadline leaving no room for receiving its response.

Also changed OpenConn to use DialContext and HandshakeContext

@martin-sucha @zimnx please let me know if this is the desired behavior, or if I missed something crucial

### DIFF
--- a/experiments/cmd/benchtab/main.go
+++ b/experiments/cmd/benchtab/main.go
@@ -44,7 +44,7 @@ func main() {
 		if err != nil {
 			log.Fatal(err)
 		}
-		initKeyspaceAndTable(initSession, config.keyspace)
+		initKeyspaceAndTable(ctx, initSession, config.keyspace)
 		initSession.Close()
 	}
 
@@ -53,21 +53,21 @@ func main() {
 	if err != nil {
 		log.Fatal(err)
 	}
-	truncateTable(session)
+	truncateTable(ctx, session)
 
 	if config.workload == Selects && !config.dontPrepare {
-		initSelectsBenchmark(session, config)
+		initSelectsBenchmark(ctx, session, config)
 	}
 
 	if config.async {
-		asyncBenchmark(&config, session)
+		asyncBenchmark(ctx, &config, session)
 	} else {
-		benchmark(&config, session)
+		benchmark(ctx, &config, session)
 	}
 }
 
 // benchmark is the same as in gocql.
-func benchmark(config *Config, session *scylla.Session) {
+func benchmark(ctx context.Context, config *Config, session *scylla.Session) {
 	var wg sync.WaitGroup
 	nextBatchStart := -config.batchSize
 
@@ -78,11 +78,11 @@ func benchmark(config *Config, session *scylla.Session) {
 		wg.Add(1)
 		go func() {
 			defer wg.Done()
-			insertQ, err := session.Prepare(insertStmt)
+			insertQ, err := session.Prepare(ctx, insertStmt)
 			if err != nil {
 				log.Fatal(err)
 			}
-			selectQ, err := session.Prepare(selectStmt)
+			selectQ, err := session.Prepare(ctx, selectStmt)
 			if err != nil {
 				log.Fatal(err)
 			}
@@ -98,7 +98,7 @@ func benchmark(config *Config, session *scylla.Session) {
 
 				for pk := curBatchStart; pk < curBatchEnd; pk++ {
 					if config.workload == Inserts || config.workload == Mixed {
-						_, err := insertQ.BindInt64(0, pk).BindInt64(1, 2*pk).BindInt64(2, 3*pk).Exec()
+						_, err := insertQ.BindInt64(0, pk).BindInt64(1, 2*pk).BindInt64(2, 3*pk).Exec(ctx)
 						if err != nil {
 							panic(err)
 						}
@@ -106,7 +106,7 @@ func benchmark(config *Config, session *scylla.Session) {
 
 					if config.workload == Selects || config.workload == Mixed {
 						var v1, v2 int64
-						res, err := selectQ.BindInt64(0, pk).Exec()
+						res, err := selectQ.BindInt64(0, pk).Exec(ctx)
 						if err != nil {
 							panic(err)
 						}
@@ -133,7 +133,7 @@ func benchmark(config *Config, session *scylla.Session) {
 	log.Printf("Finished\nBenchmark time: %d ms\n", benchTime.Milliseconds())
 }
 
-func asyncBenchmark(config *Config, session *scylla.Session) {
+func asyncBenchmark(ctx context.Context, config *Config, session *scylla.Session) {
 	var wg sync.WaitGroup
 	nextBatchStart := -config.batchSize
 
@@ -145,11 +145,11 @@ func asyncBenchmark(config *Config, session *scylla.Session) {
 		go func() {
 			defer wg.Done()
 
-			insertQ, err := session.Prepare(insertStmt)
+			insertQ, err := session.Prepare(ctx, insertStmt)
 			if err != nil {
 				log.Fatal(err)
 			}
-			selectQ, err := session.Prepare(selectStmt)
+			selectQ, err := session.Prepare(ctx, selectStmt)
 			if err != nil {
 				log.Fatal(err)
 			}
@@ -164,11 +164,11 @@ func asyncBenchmark(config *Config, session *scylla.Session) {
 				curBatchEnd := min(curBatchStart+config.batchSize, config.tasks)
 
 				if config.workload == Inserts || config.workload == Mixed {
-					asyncInserts(&insertQ, curBatchStart, curBatchEnd)
+					asyncInserts(ctx, &insertQ, curBatchStart, curBatchEnd)
 				}
 
 				if config.workload == Selects || config.workload == Mixed {
-					asyncSelects(&selectQ, curBatchStart, curBatchEnd)
+					asyncSelects(ctx, &selectQ, curBatchStart, curBatchEnd)
 				}
 			}
 		}()
@@ -179,12 +179,12 @@ func asyncBenchmark(config *Config, session *scylla.Session) {
 	log.Printf("Finished\nBenchmark time: %d ms\n", benchTime.Milliseconds())
 }
 
-func asyncInserts(insertQ *scylla.Query, curBatchStart, curBatchEnd int64) {
+func asyncInserts(ctx context.Context, insertQ *scylla.Query, curBatchStart, curBatchEnd int64) {
 	for pk := curBatchStart; pk < curBatchEnd; pk++ {
 		insertQ.BindInt64(0, pk)
 		insertQ.BindInt64(1, 2*pk)
 		insertQ.BindInt64(2, 3*pk)
-		insertQ.AsyncExec()
+		insertQ.AsyncExec(ctx)
 	}
 	for pk := curBatchStart; pk < curBatchEnd; pk++ {
 		if _, err := insertQ.Fetch(); err != nil {
@@ -193,10 +193,10 @@ func asyncInserts(insertQ *scylla.Query, curBatchStart, curBatchEnd int64) {
 	}
 }
 
-func asyncSelects(selectQ *scylla.Query, curBatchStart, curBatchEnd int64) {
+func asyncSelects(ctx context.Context, selectQ *scylla.Query, curBatchStart, curBatchEnd int64) {
 	for pk := curBatchStart; pk < curBatchEnd; pk++ {
 		selectQ.BindInt64(0, pk)
-		selectQ.AsyncExec()
+		selectQ.AsyncExec(ctx)
 	}
 	for pk := curBatchStart; pk < curBatchEnd; pk++ {
 		res, err := selectQ.Fetch()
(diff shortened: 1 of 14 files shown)


Choraden:
@Kulezi Can I get a little more **context** of what you did there?

Kulezi:
ping @martin-sucha, @zimnx

experiments/cmd/benchtab/main.go
```diff
@@ -34,7 +35,7 @@ func main() {
 	cfg.Password = config.password
 
 	if !config.dontPrepare {
-		initSession, err := scylla.NewSession(cfg)
+		initSession, err := scylla.NewSession(context.Background(), cfg)
```
zimnx, line 38: How about defining a context using `signal.Notify(context.Background(), ...)` and passing it everywhere instead passing uninterruptable context?

session_integration_bench_test.go
```diff
@@ -17,33 +18,33 @@ func BenchmarkSessionQueryIntegration(b *testing.B) {
 
 	for _, stmt := range initStmts {
 		q := session.Query(stmt)
-		if _, err := q.Exec(); err != nil {
+		if _, err := q.Exec(context.Background()); err != nil {
```
zimnx, line 21: The same comment about using signal.Notify

transport/cluster.go
```diff
@@ -493,19 +494,23 @@ func (c *Cluster) handleStatusChange(v *StatusChange) {
...
 	for {
 		select {
 		case <-c.refreshChan:
-			c.tryRefresh()
+			c.tryRefresh(ctx)
 		case <-c.reopenControlChan:
-			c.tryReopenControl()
+			c.tryReopenControl(ctx)
+		case <-ctx.Done():
+			log.Println("cluster closing due to context cancellation")
```
zimnx, line 506: libraries shouldn't log anything unless it's explicitely enabled

transport/compress.go
```diff
@@ -28,12 +30,15 @@ type compr struct {
...
 
-func (c *compr) compress(dst io.Writer, src *bytes.Buffer) (written int64, err error) {
+func (c *compr) compress(ctx context.Context, dst io.Writer, src *bytes.Buffer) (written int64, err error) {
 	c.buf = *src
 	buf := c.buf.Bytes()
 	read := len(c.buf.Bytes())
 
 	if read <= 9 {
+		if ctx.Err() != nil {
+			return 0, fmt.Errorf("request aborted due to context cancellation")
```
zimnx, line 40: context might be timed out, not cancelled so error message is wrong

transport/compress.go
```diff
@@ -28,12 +30,15 @@ type compr struct {
...
 }
 
-func (c *compr) compress(dst io.Writer, src *bytes.Buffer) (written int64, err error) {
+func (c *compr) compress(ctx context.Context, dst io.Writer, src *bytes.Buffer) (written int64, err error) {
 	c.buf = *src
 	buf := c.buf.Bytes()
 	read := len(c.buf.Bytes())
 
 	if read <= 9 {
+		if ctx.Err() != nil {
```
zimnx, line 39: wouldn't checking if context is done and returning wrapped error be better here?

transport/compress.go
```diff
@@ -45,6 +50,10 @@ func (c *compr) compress(dst io.Writer, src *bytes.Buffer) (written int64, err e
 		return 0, err
 	}
 
+	if ctx.Err() != nil {
```
zimnx, line 53: ditto

transport/conn.go
```diff
@@ -114,11 +117,20 @@ func (c *connWriter) send(r request) error {
 	l := uint32(len(b) - frame.HeaderSize)
 	binary.BigEndian.PutUint32(b[5:9], l)
 
+	if ctx.Err() != nil {
```
zimnx, line 120: ditto

transport/conn.go
```diff
@@ -171,7 +183,7 @@ func (c *connReader) handler(streamID frame.StreamID) ResponseHandler {
 	return h
 }
 
-func (c *connReader) loop() {
+func (c *connReader) loop(ctx context.Context) {
 	c.bufw = frame.BufferWriter(&c.buf)
 	for {
 		resp := c.recv()
```
zimnx, line 189: ctx is unused and looking like this loop blocks on recv, which should be interruptable

transport/conn.go
```diff
@@ -639,8 +651,11 @@ func MakeResponseHandlerWithError(err error) ResponseHandler {
 	return h
 }
 
-func (c *Conn) sendRequest(req frame.Request, compress, tracing bool) (frame.Response, error) {
-	c.sendController()
+func (c *Conn) sendRequest(ctx context.Context, req frame.Request, compress, tracing bool) (frame.Response, error) {
+	c.sendController(ctx)
```
zimnx, line 655: shouldn't `sendController` return error?

transport/conn.go
```diff
@@ -655,22 +670,29 @@ func (c *Conn) sendRequest(req frame.Request, compress, tracing bool) (frame.Res
...
 	c.w.submit(r)
 
-	resp := <-h
-
-	return resp.Response, resp.Err
+	select {
+	case resp := <-h:
+		return resp.Response, resp.Err
+	case <-ctx.Done():
+		return nil, fmt.Errorf("no response before context cancellation")
```
zimnx, line 685: context might be timed out, error message is wrong

transport/conn.go
```diff
@@ -655,22 +670,29 @@ func (c *Conn) sendRequest(req frame.Request, compress, tracing bool) (frame.Res
...
+		// TODO: timeout chan?
```
zimnx, line 686: TODO?

transport/conn.go
```diff
@@ -655,22 +670,29 @@ func (c *Conn) sendRequest(req frame.Request, compress, tracing bool) (frame.Res
...
 }
 
-func (c *Conn) asyncSendRequest(req frame.Request, compress, tracing bool, h ResponseHandler) {
+func (c *Conn) asyncSendRequest(ctx context.Context, req frame.Request, compress, tracing bool, h ResponseHandler) {
 control:
-	c.sendController()
-
+	c.sendController(ctx)
+	if ctx.Err() != nil {
+		h <- response{Err: fmt.Errorf("no response before context cancellation")}
```
zimnx, line 694: ditto

transport/conn.go
```diff
@@ -695,23 +718,23 @@ control:
...
 }
 
-func (c *Conn) sendController() {
+func (c *Conn) sendController(ctx context.Context) {
 	for {
-		if size := c.Waiting(); size < targetWaiting {
+		if size := c.Waiting(); size < targetWaiting || ctx.Err() != nil {
 			break
 		}
 		time.Sleep(50 * time.Millisecond)
```
zimnx, line 726: This should use ticker and react to context interrupt in select

transport/conn_integration_test.go
```diff
@@ -22,7 +23,7 @@ func TestOpenShardConnIntegration(t *testing.T) {
 
 	for i := uint16(0); i < si.NrShards; i++ {
 		si.Shard = i
-		c, err := OpenShardConn(TestHost+":19042", si, DefaultConnConfig(""))
+		c, err := OpenShardConn(context.Background(), TestHost+":19042", si, DefaultConnConfig(""))
```
zimnx, line 26: ditto

transport/pool.go
```diff
@@ -23,15 +24,19 @@ type ConnPool struct {
 	connObs      ConnObserver
 }
 
-func NewConnPool(host string, cfg ConnConfig) (*ConnPool, error) {
+func NewConnPool(ctx context.Context, host string, cfg ConnConfig) (*ConnPool, error) {
+	if err := ctx.Err(); err != nil {
+		return nil, err
+	}
```
zimnx, line 30: no point of checking it here, you should check it just before/during/after blocking operations

transport/pool.go
```diff
@@ -192,14 +197,14 @@ func (r *PoolRefiller) onConnClose(conn *Conn) {
 
 const fillBackoff = time.Second
 
-func (r *PoolRefiller) loop() {
-	r.fill()
+func (r *PoolRefiller) loop(ctx context.Context) {
+	r.fill(ctx)
 
 	timer := time.NewTicker(fillBackoff)
```
zimnx, line 203: nit: it's ticker, not timer

transport/pool.go
```diff
@@ -192,14 +197,14 @@ func (r *PoolRefiller) onConnClose(conn *Conn) {
...
 	for {
 		select {
```
zimnx, line 205: add case for ctx.Done(), you should interrupt the for loop when it happens

zimnx:
Please ask for review only when PR is ready to be merged. I usually don't look at the drafts at all.

transport/conn.go
Kulezi: Thanks for noticing. Actually reader should not take the context as this recv can only be interrupted by its underlying connection being closed, which is already done in cluster/pool when it sees that the context is done.

Kulezi:
Made changes with regards to @zimnx comments, also added an integration test that checks if the behavior is correct.
Opened issue #271 regarding logs, This PR does not implement any additional wait for responses that reached the driver near contexts deadline, I suppose if it's really necessary it can be added later.

transport/conn.go
```diff
@@ -171,6 +183,7 @@ func (c *connReader) handler(streamID frame.StreamID) ResponseHandler {
 	return h
 }
 
+// loop terminates when its connection gets closed by the pool, especially when session context is done.
```
zimnx, line 186: I don't see any context being used here, why there's a hidden side effect?

transport/conn.go
```diff
@@ -639,8 +652,13 @@ func MakeResponseHandlerWithError(err error) ResponseHandler {
...
 
-func (c *Conn) sendRequest(req frame.Request, compress, tracing bool) (frame.Response, error) {
-	c.sendController()
+func (c *Conn) sendRequest(ctx context.Context, req frame.Request, compress, tracing bool) (frame.Response, error) {
+	if err := c.sendController(ctx); err != nil {
+		return nil, err
+	}
+	if err := ctx.Err(); err != nil {
+		return nil, err
+	}
```
zimnx, line 661: shouldn't sendController return error when context is cancelled? If so, this check is redundant

transport/conn.go
```diff
@@ -36,6 +37,8 @@ type request struct {
 	Compress        bool
 	Tracing         bool
 	ResponseHandler ResponseHandler
+
+	ctx context.Context // nolint:containedctx // cancelling sending request can't be done without it.
```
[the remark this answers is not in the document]
Kulezi: Those request structs are sent to the writer loop which sends them to scylla. This is the only way of knowing if a single requests context is done inside the loop, allowing to skip sending it if the context is done.
Scenario: someone wants to cancel a single request but not the whole session.

transport/conn.go
Kulezi: The way reader loop works is that its blocking on io.Copy which from what I recall cannot be interrupted using contexts.

When session is being closed it closes all its connections, making io.Copy return an io error, closing the reader loop.
This occurs especially when session context is closed.

transport/conn.go
Kulezi: This check is not redundant as sendController could finish immediately without checking the ctx,
however it could return ctx.Err() instead of nil, making this check redundant (making it easier not to forget)

session_integration_test.go
```diff
@@ -403,3 +414,68 @@ func TestPrepareIntegration(t *testing.T) {
...
+
+	// Before stop queries should succeed.
+	for _, stmt := range initStmts {
+		q := session.Query(stmt)
+		if _, err := q.Exec(ctx); err != nil {
+			t.Fatal(err)
+		}
+
+		// Await schema agreement, TODO: implement true schema agreement.
+		time.Sleep(time.Second)
```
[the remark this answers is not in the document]
Kulezi: Issue is already there, https://github.com/scylladb/scylla-go-driver/issues/252, added the mention about removal.

Kulezi:
Applied changes suggested by @zimnx.

I also noticed that skipping a single request can increase the inflight request counter in connWriter.loop() even though it wasn't sent, fixed this by changing the behavior from returning nil to returning a custom error in this case. This way the end user will also know the reason request was skipped, and unwrap the underlying context error.

Added missing propagation of request context to compress, now it checks its error too skipping after compressing and before sending.

Added missing context.Background() to _connCloseRequest.

transport/conn.go
```diff
@@ -81,9 +96,12 @@ func (c *connWriter) loop() {
 				return
 			}
 			c.stats.inQueue.Dec()
-			if err := c.send(r); err != nil {
-				log.Printf("%s fatal send error, closing connection due to %s", c.connString(), err)
+			if err := c.send(ctx, r); err != nil {
 				r.ResponseHandler <- response{Err: fmt.Errorf("%s send: %w", c.connString(), err)}
+				if _, ok := err.(*skippedError); ok {
+					continue
```
Choraden, line 102: Wouldn't it casue streamIDs leak?

Choraden: ConnWriter should be able to free streamID. Some kind of a callback with pointer to connReader seems to be convenient solution, but I'm not sure if it is the best.

Kulezi: Good point, this is how i'd like to resolve it, as in sendRequest we can't be sure we encounter a skippedError before looking at ctx.Done().

Kulezi:
Fixed the StreamID leak, and made the commit messages describe the changes in a more verbose way.

zimnx (approved):
lgtm, thanks

PR #265, opened 2022-07-26
Support for general binding and parsing any types

Kulezi:
- Added support for binding any type with use of gocql.Marshal
- Added support for scanning any type with use of gocql.Unmarshal
- Binding to unprepared queries can now be done only with type-specialized methods
- Incorrect binding will now return an error on query execution
- Added benchmarks comparing conversions using general and concrete methods, results:
```    
goos: linux
goarch: amd64
pkg: github.com/mmatczuk/scylla-go-driver
cpu: 11th Gen Intel(R) Core(TM) i7-11850H @ 2.50GHz
BenchmarkBindConcrete-16        241682629                5.023 ns/op           0 B/op          0 allocs/op
BenchmarkBindAny-16             29523056                39.13 ns/op           15 B/op          1 allocs/op
BenchmarkAsConcrete-16          511050747                2.350 ns/op           0 B/op          0 allocs/op
BenchmarkAsAny-16               52738100                21.69 ns/op            8 B/op          1 allocs/op
PASS
ok      github.com/mmatczuk/scylla-go-driver    5.528s
```

I think this will be needed to make a compatibility layer for gocql and gocqlX, benchmarks show that it's much slower than concrete bindings, however I can barely see any difference with `make run-benchtab`, maybe because of the driver having spare time due to waiting on syscalls.


Kulezi:
This will be used in a separate package providing gocql compatibility.

PR #267, opened 2022-08-05
query: use retry policy

Kulezi:
The goal of this PR is to make normal and paged queries use the retry policy implented in #225.

RetryDecider implemented in #225 provides a Decide() method that for a error response will return a decision from list:
- DontRetry
- RetrySameNode
- RetryNextNode

In this PR I changed query.Exec() and query.Iter() to use this retry decider to iterate over the node plan generated by
the HostSelectionPolicy. 

The goal of the integration test implemented in this PR is to check if the actions performed by the driver match the Retry Policy decisions, and if using the Default Retry Policy produces the right amount of retries.

In the case of normal queries, a RetryDecider is allocated after the first try fails, avoiding unnecessary allocation in most cases.
In paged queries I added the needed fields to iterWorker, and reused the RetryDecider by calling it's Reset() method.

In case of AsyncQuery/Exec it's unclear how it should be done, as it would either require handling retries on the side of connection / cluster which doesn't make sense to me. Other possibility would be to do retries in Fetch() after reading the result, but here we would need to somehow keep the connection/node that the request was sent to to do a same node retry, etc. So for now I am leaving it as it is.

### DIFF
--- a/query.go
+++ b/query.go
@@ -201,16 +201,14 @@ func (q *Query) Iter(ctx context.Context) Iter {
 		return it
 	}
 
-	conn, err := q.pickConn(info)
-	if err != nil {
-		it.errCh <- err
-		return it
-	}
-
 	worker := iterWorker{
-		stmt:      q.stmt.Clone(),
-		conn:      conn,
+		stmt: q.stmt.Clone(),
+
+		rd:        q.session.cfg.RetryPolicy.NewRetryDecider(),
+		queryInfo: info,
+		pickNode:  q.session.cfg.HostSelectionPolicy.Node,
 		queryExec: q.exec,
+
 		requestCh: it.requestCh,
 		nextCh:    it.nextCh,
 		errCh:     it.errCh,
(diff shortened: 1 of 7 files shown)


avelanarius:
@Kulezi please rewrite the PR's description so it's more clear what was the previous behavior and exactly what was the goal of this PR. Please fix linguistical errors ("This pull request makes queries makes normal", "[...]  is allocated after first try fails avoiding").

query.go
```diff
@@ -26,20 +31,10 @@ func (q *Query) Exec() (Result, error) {
...
 }
 
-func (q *Query) pickConn() (*transport.Conn, error) {
-	token, tokenAware := q.token()
-	info, err := q.info(token, tokenAware)
-	if err != nil {
-		return nil, err
-	}
-	n := q.session.cfg.HostSelectionPolicy.Node(info, 0)
+func (q *Query) pickConn(qi transport.QueryInfo) (*transport.Conn, error) {
```
avelanarius, line 65: Nit: Rewrite commit message "session: pickConn, take QueryInfo as argument instead of calculating it" to something better like: "session: take QueryInfo in pickConn and node.Conn()".

avelanarius:
In "session: pickConn, take QueryInfo as argument instead of calculating it" commit message there is "removing the need to export queryInfo fields" - could you clarify what you mean by "removing the need to export queryInfo fields"?

Kulezi:
> In "session: pickConn, take QueryInfo as argument instead of calculating it" commit message there is "removing the need to export queryInfo fields" - could you clarify what you mean by "removing the need to export queryInfo fields"?

This commit message is a product of a badly done squash, before this was done in two commits, first would export Token and TokenAware fields so they would be visible outside of transport package but changes in the second commit allowed not to do that, so I squashed them, but I the message by mistake.

avelanarius (approved):
LGTM

PR #268, opened 2022-08-07
Replace mmatczuk with scylladb in module path

martin-sucha:
The repository was moved from

https://github.com/mmatczuk/scylla-go-driver

to

https://github.com/scylladb/scylla-go-driver

but the module path still contained the old name.

While it is possible to import the driver with the old import path,
using the one with scylladb will make importing the driver
a little bit more obvious.

The change might break external modules that depend on the driver
now, but I don't expect there to be many of those, since the
driver is still experimental.

### DIFF
--- a/transport/policy_test.go
+++ b/transport/policy_test.go
@@ -3,7 +3,7 @@ package transport
 import (
 	"testing"
 
-	"github.com/mmatczuk/scylla-go-driver/frame"
+	"github.com/scylladb/scylla-go-driver/frame"
 )
 
 // Round-Robin tests can't be run in parallel because
--- a/errors.go
+++ b/errors.go
@@ -1,6 +1,6 @@
 package scylla
 
-import "github.com/mmatczuk/scylla-go-driver/frame"
+import "github.com/scylladb/scylla-go-driver/frame"
 
 // This file is identical to errors.go located in scylla pkg.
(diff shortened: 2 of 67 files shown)


transport/policy_test.go
```diff
@@ -239,24 +239,24 @@ func TestTokenAwareSimpleStrategyPolicy(t *testing.T) { //nolint:paralleltest //
...
+Datacenter:       her
+nodes in rack r3: 5 6
+nodes in rack r4: 7 8
 
-	Keyspace:         "waw/her"
-	strategy: network topology
-	replication factors: waw: 2 her: 3
+Keyspace:         "waw/her"
+strategy: network topology
+replication factors: waw: 2 her: 3
```
Kulezi, line 259: Is this intentional? This doesn't need a change.

Kulezi: Rest LGTM

Kulezi (changes requested):
Please restore indentation in policy_test.go so linter does not complain, then it will be merged.

transport/policy_test.go
martin-sucha: Indeed this change was not intentional. Thanks for noticing! Fixed.

However, you might want to apply it in a separate pull request anyway, it is a result of running `go fmt` (version 1.19) on the file.

Kulezi (approved):

Issue #271, opened 2022-08-10
session: add log configuration


Kulezi:
Right now the driver logs important events to stderr, it should be off by default and configurable by the user through sessionConfig.

Maybe using an interface like gocql: https://github.com/gocql/gocql/blob/v1.2.0/logger.go#L9
or https://github.com/scylladb/go-log

martin-sucha:
I'd prefer some kind of interface where I can plug a custom implementation. We have an internal logging package (based on zap) that always adds some custom fields.

The driver should also log query warnings by default to the configured logger and allow users to explicitly turn the warnings off if needed. That would make warnings more useful. With gocql probably most people ignore them. (but maybe this is for a separate issue?)

PR #276, opened 2022-08-26
Run integration tests in Github Workflows

Kulezi:
The goal of this PR is to run integration tests on the cluster described by docker-compose.yml in GH workflows.
It extends the Makefile with ability to run integration tests in docker, this is used by the modified workflow.

This PR also resolves a small bug in OpenShardConn, when context is done, instead of immediate return from the function, it would try all ports as connections wouldn't open due to context being done, bloating the integration test log with thousands of lines.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -9,6 +9,8 @@ on:
   pull_request:
 permissions:
   contents: read
+env:
+  SCYLLA_ARGS: "--seeds=node1,node2 --authenticator PasswordAuthenticator --api-address 0.0.0.0"
 jobs:
   sanity:
     name: Sanity check
@@ -28,6 +30,21 @@ jobs:
         with:
           version: v1.46.2
 
+      - name: Setup 3-node Scylla cluster
+        run: |
+          sudo sh -c "echo 2097152 >> /proc/sys/fs/aio-max-nr"
+          docker-compose up -d
+
       - name: Run tests
         run: make test
 
+      - name: Ensure nodes are up
+        run: |
+          until docker compose exec node1 nodetool status; do sleep 1; done
+          while [[ $(docker compose exec node1 nodetool status | grep -c ^UN) != "3" ]]; do sleep 1; done
+
+      - name: Run integration tests
+        run: make docker-integration-test
+        
+      - name: Stop cluster
+        run: docker-compose down
--- a/Makefile
+++ b/Makefile
@@ -58,6 +58,16 @@ else
 	$(error Unsupported OS $(OS))
 endif
 
+.PHONY: docker-integration-test
+docker-integration-test: RUN=Integration
+docker-integration-test:
+	@CGO_ENABLED=0 GOOS=linux go test -v -tags integration -c -o ./integration-test.dev $(PKG)
+	@docker run --name "integration-test" \
+		--network scylla_go_driver_public \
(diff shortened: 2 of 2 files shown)


Makefile
```diff
@@ -57,6 +57,16 @@ else
...
 
+.PHONY: docker-integration-test
+docker-integration-test: RUN=Integration
+docker-integration-test:
+	@CGO_ENABLED=0 GOOS=linux go test -v -tags integration -c -o ./integration-test.dev $(PKG)
+	@docker run --name "integration-test" \
+		--network scylla_go_driver_public \
+		-v "$(PWD)/testdata:/testdata" \
+		-v "$(PWD)/integration-test.dev:/usr/bin/integration-test:ro" \
+		-i --read-only --rm ubuntu integration-test -test.v -test.run $(RUN) $(ARGS)
```
piodul, line 68: Why is `testdata/testdata`, mounted here, but in the `pkg-integration-test` target it's not?

transport/conn.go
```diff
@@ -376,6 +376,10 @@ func OpenShardConn(ctx context.Context, addr string, si ShardInfo, cfg ConnConfi
 	it := ShardPortIterator(si)
 	maxTries := (maxPort-minPort+1)/int(si.NrShards) + 1
 	for i := 0; i < maxTries; i++ {
+		if err := ctx.Err(); err != nil {
+			return nil, fmt.Errorf("aborted opening connection on shard %d: %w", si.Shard, err)
+		}
+
```
piodul, line 382: The fix this commit introduces doesn't seem to be related to the main purpose of the PR (adding integration tests to the CI). It should be sent in a separate PR.

.github/workflows/go.yml
```diff
@@ -28,6 +30,21 @@ jobs:
...
+      - name: Ensure nodes are up
+        run: |
+          until docker compose exec node1 nodetool status; do sleep 1; done
+          while [[ $(docker compose exec node1 nodetool status | grep -c ^UN) != "3" ]]; do sleep 1; done
+
+      - name: Run integration tests
+        run: make docker-integration-test
+        
+      - name: Stop cluster
+        run: GITHUB_WORKSPACE=${{ github.workspace }} docker-compose down
```
piodul, line 50: What is the purpose for the `GITHUB_WORKSPACE` env variable here?

Makefile
Kulezi: Missed this as workflow is being ran on ubuntu not mac, fixed.

transport/conn.go
Kulezi: Will move that to #271 as its related to logs, dropped

.github/workflows/go.yml
Kulezi: It's unused, removed.

Makefile
```diff
@@ -51,12 +51,23 @@ else ifeq ($(OS),Darwin)
 	@CGO_ENABLED=0 GOOS=linux go test -v -tags integration -c -o ./integration-test.dev $(PKG)
 	@docker run --name "integration-test" \
 		--network scylla_go_driver_public \
+		-v "$(PWD)/testdata:/testdata" \
```
piodul, line 54: Nit: it would be preferable to fix this in a separate commit. The current commit is about the `docker-integration-test`, so this change seems unrelated.

piodul (approved):

PR #278, opened 2022-09-06
session: allow log configuration using Logger interface.

### LINKED ISSUE #271: session: add log configuration

Kulezi:
This PR provides a Logger interface allowing to supply
a custom logger as part of (Session)ConnConfig.

It also provides three implementations of Logger:
- DefaultLogger, logging only warnings
- DebugLogger, logging everything that can be useful for debugging,
  it is used by default in tests
- NopLogger, logging nothing

Fixes #271

### DIFF
--- a/session.go
+++ b/session.go
@@ -3,7 +3,6 @@ package scylla
 import (
 	"context"
 	"fmt"
-	"log"
 	"sync"
 	"time"
 
@@ -271,6 +270,6 @@ func (s *Session) NewTokenAwareDCAwarePolicy(localDC string) transport.HostSelec
 }
 
 func (s *Session) Close() {
-	log.Println("session: close")
+	s.cfg.Logger.Info("session: close")
 	s.cluster.Close()
 }
--- a/transport/conn.go
+++ b/transport/conn.go
@@ -8,7 +8,6 @@ import (
 	"errors"
 	"fmt"
 	"io"
-	"log"
 	"net"
 	"strconv"
 	"strings"
@@ -19,6 +18,7 @@ import (
 	"github.com/scylladb/scylla-go-driver/frame"
 	. "github.com/scylladb/scylla-go-driver/frame/request"
 	. "github.com/scylladb/scylla-go-driver/frame/response"
+	"github.com/scylladb/scylla-go-driver/log"
 
 	"go.uber.org/atomic"
 )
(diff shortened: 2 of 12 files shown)


transport/logger.go
```diff
@@ -0,0 +1,69 @@
+package transport
```
zimnx, line 1: how does logger relates to transport? It feels like it belongs to root package

Kulezi: Almost all logging takes place in the transport package, and putting the entire logger inside of scylla package creates an import cycle, transport package should not have dependencies in scylla package.

I suppose Logger interface could be declared identically in both transport and scylla packages instead of aliasing it in scylla package so it shows up as the whole declaration instead of an alias in docs.

It also could have its own package but that seems excessive to me.

transport/conn.go
```diff
@@ -383,6 +383,10 @@ func OpenShardConn(ctx context.Context, addr string, si ShardInfo, cfg ConnConfi
 	it := ShardPortIterator(si)
 	maxTries := (maxPort-minPort+1)/int(si.NrShards) + 1
 	for i := 0; i < maxTries; i++ {
+		if err := ctx.Err(); err != nil {
+			return nil, fmt.Errorf("aborted opening connection on shard %d: %w", si.Shard, err)
+		}
+
```
piodul, line 389: Does the second commit (`transport: conn, abort OpenShardConn faster when context is done`) belong to this PR? It looks like it can be submitted separately and reviewed independently.

Kulezi: Moved to #286.

logger/logger.go
```diff
@@ -0,0 +1,69 @@
+package logger
```
zimnx, line 1:
```suggestion
package log
```
"log" is in sync with other logging packages

session.go
```diff
@@ -66,6 +66,18 @@ var (
...
+}
+
+func NewDebugLogger() log.Logger {
+	return log.NewDebugLogger()
+}
+
+func NewNopLogger() log.Logger {
+	return log.NopLogger{}
+}
+
```
zimnx, line 80: These shouldn't be exposed in root package, users aren't generally interested such details, default session should use default logger. Advanced users who would like to change logging details, can import log package.

Kulezi:
Rebased, renamed the new logging package to log. Integration tests don't fail anymore after #283 was merged. 

Please let me know if I can merge this @zimnx.

zimnx (approved):
LGTM, thanks

Issue #279, opened 2022-09-07
cluster: nodes that were down but got up never open their pools


Kulezi:
What happens:
Driver on startup sees 3 nodes, it fails to open a connection pool to a node that is down marking it as such.
However after making all 3 nodes work, it doesn't try to open it again. 

To reproduce:
```
docker-compose up -d
docker stop scylla_go_driver-node-2
connect with the driver to the cluster
docker-compose up -d
```

Issue #222, opened 2022-04-27
Transport: nodes that are down cause nil pointer dereference 


Choraden:
Calling LeastBusyConn on a Node that is down results in panic! 
It happens when for the first time node is added to known topology in Cluster::refreshTopology()

Kulezi:
Probably caused by the same issue, found running integration tests.
```
 --- FAIL: TestContextsIntegration (0.26s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x6334f0]

goroutine 3364 [running]:
testing.tRunner.func1.2({0x674980, 0x889e30})
	/opt/hostedtoolcache/go/1.19.0/x64/src/testing/testing.go:1396 +0x24e
testing.tRunner.func1()
	/opt/hostedtoolcache/go/1.19.0/x64/src/testing/testing.go:1399 +0x39f
panic({0x674980, 0x889e30})
	/opt/hostedtoolcache/go/1.19.0/x64/src/runtime/panic.go:884 +0x212
github.com/scylladb/scylla-go-driver/transport.(*Conn).Waiting(...)
	/home/runner/work/scylla-go-driver/scylla-go-driver/transport/conn.go:777
github.com/scylladb/scylla-go-driver/transport.(*Conn).sendController(0x162f40?, {0x72a408?, 0xc000098800?})
	/home/runner/work/scylla-go-driver/scylla-go-driver/transport/conn.go:747 +0x50
github.com/scylladb/scylla-go-driver/transport.(*Conn).sendRequest(0x0, {0x72a408, 0xc000098800}, {0x729638, 0xc000140510}, 0x0, 0x0)
	/home/runner/work/scylla-go-driver/scylla-go-driver/transport/conn.go:682 +0x66
github.com/scylladb/scylla-go-driver/transport.(*Conn).Execute(_, {_, _}, {{0xc0009eaaf0, 0x10, 0x10}, {0x6bdafb, 0x25}, {0xc000998440, 0x1, ...}, ...}, ...)
	/home/runner/work/scylla-go-driver/scylla-go-driver/transport/conn.go:647 +0x235
[...]

PR #283, opened 2022-09-08
transport: avoid nil pointer dereferencing

Kulezi:
The goal of this PR is to resolve two intertwined issues:
-  #279 
The problem here was that on initial discovery of a node, if it was down it's pool couldn't be initialized, leaving it as nil.
When a topology refresh happens, the driver was reusing the pool without checking if it actually exists.<br>
For this problem the proposed solution is to move the node's connection pool and status initialization to the Node, instead of managing it inside cluster, calling node.Init() in places where the node could have status UP to ensure we have working connections to this node.

-  #222 
This issue comes with two problems: pool being nil from reasons mentioned above, and pool returning nil connections, as the node may be down, or a shard connection is missing.<br>
The proposed solution is to indicate the connection is missing by returning an error giving context about why it's missing, especially when a node's status is DOWN, making the retry policy skip nodes which don't have a working connection available.

Motivation behind this PR is to stop the integration tests from failing on GH Actions, not fix node status handling entirely, that should be done in a separate PR resolving #293 

#222 was causing panics in TestContextIntegration when connections were closed but session didn't fully close yet, running the tests in github actions confirms that this is not the case anymore.

I also ran manual tests with steps mentioned in #279, confirming that pool is initialized properly.

Fixes: #279 
Fixes: #222

### DIFF
--- a/transport/pool.go
+++ b/transport/pool.go
@@ -41,13 +41,13 @@ func (p *ConnPool) String() string {
 	return fmt.Sprintf("pool %s [shards=%d]", p.host, p.nrShards)
 }
 
-func (p *ConnPool) Conn(token Token) *Conn {
+func (p *ConnPool) Conn(token Token) (*Conn, error) {
 	idx := p.shardOf(token)
 	if conn := p.loadConn(idx); conn != nil {
 		if isHeavyLoaded(conn) {
-			return p.maybeReplaceWithLessBusyConn(conn)
+			return p.maybeReplaceWithLessBusyConn(conn), nil
 		}
-		return conn
+		return conn, nil
 	}
 	return p.LeastBusyConn()
 }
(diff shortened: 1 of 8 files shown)


transport/pool.go
```diff
@@ -80,7 +80,11 @@ func (p *ConnPool) LeastBusyConn() *Conn {
 			}
 		}
 	}
-	return leastBusyConn
+
+	if leastBusyConn == nil {
+		return nil, fmt.Errorf("no connections available for host %s", p.host)
```
piodul, line 85: Do I understand correctly that, when the node goes down, the pool becomes empty and this error will be returned a lot? If so, then maybe we could also have a global instance for this error?

piodul (changes requested):
Two things:

- The PR fixes some bugs, but I can't see any automatic tests or mentions of manual tests. If the current testing framework is powerful enough, then please add automatic regression tests. If not, then please at least test it manually and describe what has been done in the PR description.
- If the PR fixes some issues, then please add a `Fixes: #XYZ` line to the description for each of the issue fixed.

transport/pool.go
Kulezi: I think this could depend on how #293 is handled, for example if pool losing all connections would automatically mark the node down, this error would only be called in the time between the last conn dying and status being marked.

I'd keep it as simple as it is right now.

Kulezi:
Updated the description with testing info, and cleaned up mentioned errors.

piodul (approved):

PR #284, opened 2022-09-09
[WIP] Pp/go 1.19 patch

Kulezi:


transport/cluster.go
```diff
@@ -443,7 +443,7 @@ func parseTokensFromRow(n *Node, r frame.Row, ring *Ring) error {
 }
 
 func (c *Cluster) Topology() *topology {
```
martin-sucha, line 445: Is `Topology()` method still necessary? It is an exported method returning non-exported type, so I wonder if we could just remove it in favor of using `c.topology.Load()` directly.

Kulezi: `Topology()` is used in session to get all nodes to prepare a query on them. Nodes are the only field of topology that is currently exported and used outside of `transport`, so I suppose it could be removed.

The other thing is that topology is passed inside `QueryInfo` to host selection policies, but same as all fields of `QueryInfo`, it is unexported. It likely should be accessible outside of `transport`, opened #287 about it, it may influence if topology will be exported.

Issue #281, opened 2022-09-07
conn: OpenShardConn makes unnecessary retries


Kulezi:
OpenShardConn should retry dial on other local port if the one it chose is busy.
However what happens is that it retries even when it got the port but the destination was unreachable / timeouted etc. bloating the logs unnecessarily.

PR #286, opened 2022-09-12
transport: conn, abort OpenShardConn faster

Kulezi:
Currently OpenShardConn retries every time it encounters an error, regardless of whether it is context related or not, bloating the test log in case of TestContextsIntegration.

This PR changes the behavior of OpenShardConn, so it checks if context is done before trying to open a new connection, leaving immediately if it's done with appropriate error message.

Relates to #281, solving the context case, another PR will be made resolving the case of whether source port was busy or the error was something else.

### DIFF
--- a/transport/conn.go
+++ b/transport/conn.go
@@ -376,6 +376,10 @@ func OpenShardConn(ctx context.Context, addr string, si ShardInfo, cfg ConnConfi
 	it := ShardPortIterator(si)
 	maxTries := (maxPort-minPort+1)/int(si.NrShards) + 1
 	for i := 0; i < maxTries; i++ {
+		if err := ctx.Err(); err != nil {
+			return nil, fmt.Errorf("aborted opening connection on shard %d: %w", si.Shard, err)
+		}
+
 		conn, err := OpenLocalPortConn(ctx, addr, it(), cfg)
 		if err != nil {
 			log.Printf("%s dial error: %s (try %d/%d)", addr, err, i, maxTries)


piodul (commented):
I think that something different should be done. In general, we should return every error, unless it's either EADDRINUSE or EPERM. For other errors it doesn't really make sense to retry - for example, in case of a timeout it does not make sense because a different port is not better than any other.

Something like that is already done in gocql, see here: https://github.com/scylladb/gocql/blob/e3e27db4440c6a1f0029322f2bb0e266589d9f0f/scylla.go#L631-L633

I included EPERM because it turns out that on Windows some programs with administrator privileges can reserve ports for themselves and Windows returns a WinApi equivalent of EPERM if you try to use it; we had an issue like that in the rust driver: https://github.com/scylladb/scylla-rust-driver/issues/205

Kulezi:
> I think that something different should be done. In general, we should return every error, unless it's either EADDRINUSE or EPERM. For other errors it doesn't really make sense to retry - for example, in case of a timeout it does not make sense because a different port is not better than any other.
> 
> Something like that is already done in gocql, see here: https://github.com/scylladb/gocql/blob/e3e27db4440c6a1f0029322f2bb0e266589d9f0f/scylla.go#L631-L633
> 
> I included EPERM because it turns out that on Windows some programs with administrator privileges can reserve ports for themselves and Windows returns a WinApi equivalent of EPERM if you try to use it; we had an issue like that in the rust driver: [scylladb/scylla-rust-driver#205](https://github.com/scylladb/scylla-rust-driver/issues/205)

Implemented that as part of #295 bugfix PR. Closing this one its favor.

